PR #2047, opened 2014-12-03
Charass delete

bendiy:
This is an ASSET customization I think we can move to core.


gilmoskowitz:
+1 but reopen against 4_9_x?

bendiy:
Opened PR on 4_9_x:
https://github.com/xtuple/xtuple/pull/2108

PR #2082, opened 2014-12-16
Email integration

shackbarth:
Not ready for pull yet, and destined for 4.9, but worthwhile opening for discussion.

This code exposes a REST endpoint to show all messages to/from a particular email address. The way the Qt client would consume this endpoint would be that there'd be an email tab in the contact workspace (something similar could be set up for accounts, too) that would make a HTTP request to this endpoint, and would render the resultant JSON however it sees fit.

Configuration options: 

1. The simplest way to set this up is to go into your mail server, add an archive-y email address, bcc all communication to that address, and then enter into our system the creds for that address. The imap search can be done in real time on that address without the need to put anything in a table in our database.

2. Option 1 probably won't work for gmail. It's also possible to put in an array of imap creds, for each user, and the search will look across all configured imap accounts. In this configuration option (and the third, it's furthermore possible to specify a given folder ("INBOX.ERP") as the place the search will look. This is a good option for users that are nervous about having *every single* one of their emails in the ERP.

3. It's furthermore possible to run these searches on a schedule and cache the results into the database. This would be an improvement over option 2 because it would work even if users delete their emails or if their account is removed. The downside is that you're likely to end up with a bloated database.

### DIFF
--- a/node-datasource/controllers/ImapController.js
+++ b/node-datasource/controllers/ImapController.js
@@ -7,6 +7,8 @@
     var Imap = require('imap'),
       inspect = require('util').inspect,
       returnResults = [];
+    var MailParser = require("mailparser").MailParser,
+      mailparser = new MailParser();
 
     var imap = new Imap(X.options.datasource.imapUser);
 
@@ -19,7 +21,17 @@
         if (err) throw err;
         imap.search([ 'ALL', ['FROM', req.query.address] ], function (err, results) {
           if (err) throw err;
-          var f = imap.fetch(results, { bodies: ['HEADER.FIELDS (FROM TO)','TEXT'] });
+          console.log("res count", results.length);
+          mailparser.on("end", function (mail_object){
+            returnResults.push(mail_object);
+            console.log("mailparser end");
+            if (returnResults.length === results.length) {
+              res.send(returnResults);
+            }
+          });
+
+          var f = imap.fetch(results, { bodies: [''] });
+
           f.on('message', function (msg, seqno) {
             //console.log('Message #%d', seqno);
             var prefix = '(#' + seqno + ') ';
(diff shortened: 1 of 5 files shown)


gilmoskowitz:
Interesting proof of concept. The code looks fairly simple to maintain. Open questions:

- Where should the email credentials be stored?
- If the list of mailboxes != all, what is a reasonable way to maintain this list?
- Which of the three options should we pursue further?
  - 3 is closest to what we do now with some advantages and disadvantages 
  - 2 is appealing except for the problems with deleted messages and accounts

Whichever way we go we would have to create a UI to enter & manage credentials and mailbox lists, then centralize storage of these data. The Trojita IMAP client provides that UI now. These data are stored as Qt properties on the end user machine and are read by a separate sync process that is started on xTuple client startup.

nedlilly:
Yes, we'd want to maintain the ability for users to select folders to be synced through the UI.

Sure would be nice to a) have uncomplicated gmail support, and b) support Sent mail too (the biggest functional hole in the current solution, imho)

This sounds cool, Steve.

bendiy:
@shackbarth I see your running this through `require('connect-ensure-login').ensureLoggedIn({redirectTo: "/logout"}`. How are we going to authenticate the Qt Client's browser so it will have a session and can make these requests to the REST API?

shackbarth:
@gilmoskowitz: 
-Right now you store the STMP creds in the config.js file. This is going to have to move to the database by the time we release, and we'll have to write a (very simple!) setup screen in Qt. (Unless the customer wants configuration option 1, in which case it'd probably be fine just to put the one archivey-email creds into the config.)
-Right now you store the mailbox name(s) to sync in the config file. This would also be moved to the same screen. Although, could we just say that the convention is that users should set up a folder called ERP and everything in there will be synched? That's what's written into `sample_config`.

shackbarth:
@bendiy: That's an open question for any possible web embedding in Qt, such as dashboard-anything. The best answer I have right now is, Qt should log in as the browser to the datasource at the moment of login, and keepalive the session until logout. I'm open to other ideas, though. I don't want to save the user's password in state, obviously.

PR #2090, opened 2014-12-19
Issue #25035:add prj_number and better parent order info

garyhgohoos:


foundation-database/public/functions/formatwonumber.sql
```diff
@@ -1,16 +1,32 @@
...
+                                          pType TEXT) RETURNS TEXT IMMUTABLE AS $$
+-- Copyright (c) 1999-2014 by OpenMFG LLC, d/b/a xTuple. 
+-- See www.xtuple.com/CPAL for the full text of the software license.
+
+BEGIN
+
+  IF (pType = 'womatl') THEN
+    RETURN ( SELECT (wo_number::TEXT || '-' || wo_subnumber::TEXT)
+             FROM womatl JOIN wo ON (wo_id=womatl_id)
+             WHERE (wo_id=pId) );
```
gilmoskowitz: Since the `where` criterion is the same in both queries, what's the benefit of adding the `pType` overload to `formatWoNumber`? Is it explicitly to force blank output if there is no work order material?

PR #2102, opened 2015-01-05
Copy Accounting Year (and corresponding periods) to a new fiscal year

andersdd:
Db logic.  Also has qt-client changes required.


shackbarth:
Is there an incident for this, and does it have anything to do with 
https://github.com/xtuple/xtuple/pull/2118
?

andersdd:
No incident.  David B and others were debating this on Slack and I
already had most of the logic done from elsewhere so decided to put it
in place.

ssshyam:
Unable to perform UAT: This PR depends on qt-client PR #576

foundation-database/public/functions/copyaccountingyearperiod.sql
```diff
@@ -0,0 +1,45 @@
...
+BEGIN
+
+  -- First create new fiscal year
+  INSERT INTO yearperiod (yearperiod_start, yearperiod_end, yearperiod_closed)
+    SELECT yearperiod_start + INTERVAL ' 1 year',  yearperiod_end + INTERVAL ' 1 year', FALSE
+      FROM yearperiod
+      WHERE yearperiod_id = pYearPeriodid
+  RETURNING yearperiod_id INTO _newYear;
+
+  IF (NOT _newYear > 0) THEN
```
gilmoskowitz: If the insert fails then `_newYear` will be `null` and this `IF` will evaluate to false, so the exception won't be raised (welcome to NULL hell). It's safer to check `IF _newYear IS NULL`.

foundation-database/public/functions/copyaccountingyearperiod.sql
```diff
@@ -0,0 +1,45 @@
...
+    RAISE EXCEPTION 'An error occurred creating the Fiscal Year';
+    RETURN -1;
```
gilmoskowitz: The `return` won't get executed. Better to include the `-1` in the message itself so the client can translate the message. See the [exception example](https://github.com/xtuple/qt-client/wiki/xTuple-Desktop-Client-Coding-Standards#plpgsql), at least until we redesign error handling again.

PR #2108, opened 2015-01-07
Charass delete

bendiy:
See: https://github.com/xtuple/xtuple/pull/2047#issuecomment-68724985


harikajulaganti:
Performed UAT : Observed that on deleting anything that a Characteristic is attached to, the Characteristic entry get deleted from 'charass' table . Issue is fixed

PR #2119, opened 2015-01-14
Expose user preferences to REST

shackbarth:
Easy way to persist and retrieve user preferences via REST. The user is assumed to be whatever user is making the request. Note that we persist to `xt.userpref` and not `public.usrpref`. 

```
$.ajax({
      url: '/' + org + '/browser-api/v1/services/user-preference/commit-preference',
      type: "POST",
      dataType: "json",
      data: {
        attributes: [
          "my_userpref_name",
          "my_userpref_value"
        ]
      }
});
```

```
$.ajax({
      url: '/' + org + '/browser-api/v1/services/user-preference/get-preference',
      type: "POST",
      dataType: "json",
      data: {
        attributes: [
          "my_userpref_value"
        ]
      }
});
```


shackbarth:
@bendiy I'm not sure what all the annotations are for, but this does work in my local testing. Anything convention I'm stepping on here?

PR #2124, opened 2015-01-15
add dashboard-anything to default extension list

shackbarth:
No need to pull this until dashboard-anything is more ready, but this is all that's necessary to put dashboard-anything into new installations.

To get it onto existing installs, execute this command (even though that directory won't exist yet):

```
$ ./scripts/build_app.js -d demo_dev -e node_modules/xtuple-dashboard-anything
```


shackbarth:
The page itself will then be accessible at 
`https://{host}/{org}/npm/xtuple-dashboard-anything/public/index.html`

You'll want to have logged in to the webapp first, which I'm assuming we're going to build into the Qt login process.

shackbarth:
Test failure: need to install grunt-cli in xtuple-dashboard-anything, or possibly reconfigure to just use grunt.

PR #2127, opened 2015-01-19
Issue #25226:return immediately if no inv history to thaw

garyhgohoos:


Panduk:
Performed UAT: Observed the mentioned changes in 'thawitemsite.sql'

PR #2129, opened 2015-01-19
Issue #25195 Forward Update Trial Balances when creating new Fiscal Periods

andersdd:
Rolls forward from the latest Trial Balance entry so performance does not appear too bad.

### DIFF
--- a/foundation-database/public/functions/createaccountingperiod.sql
+++ b/foundation-database/public/functions/createaccountingperiod.sql
@@ -91,6 +91,13 @@ BEGIN
     PERFORM postIntoTrialBalance(_r.gltrans_sequence);
   END LOOP;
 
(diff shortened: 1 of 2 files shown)


ssshyam:
Performed UAT: Issue is fixed

andersdd:
Issue raised by @ssshyam was caused by the Accounting Period deletion not deleting the Trial Balance entries for the Accounting Period.  Essentially, deleting the Accounting Period left orphaned entries in the trialbal table.

Added delete of Trialbalance entries when deleting accounting period and retested forward updating when re-creating the new period.

gpazo:
@ssshyam can you perform UAT on his updated changes? If not, we can merge and test in the upcoming beta.

code review: +1 looks good

ssshyam:
Performed UAT: Issue is fixed

PR #2137, opened 2015-01-22
add shackbarth's utility

gilmoskowitz:


shackbarth:
This is already here:
https://github.com/xtuple/private-extensions/tree/4_9_x/source/incident_plus/scripts
although that version may be a bit out of date. I'd rather keep it out of our core.

PR #2147, opened 2015-01-26
Issue #25259:revert change by Issue 24698

garyhgohoos:


foundation-database/public/views/invoiceitem.sql
```diff
@@ -12,8 +12,7 @@ SELECT invcitem.*, itemsite_id, cohead_number,
        ( SELECT COALESCE(SUM(taxhist_tax), 0)
          FROM invcitemtax
          WHERE (taxhist_parent_id = invcitem_id) ) AS tax,
-       ( SELECT COALESCE((coitem_unitcost * invcitem_billed),
-                         SUM(shipitem_value),
+       ( SELECT COALESCE(SUM(shipitem_value),
```
gilmoskowitz: So do we re-open 24698? How do we fix 24698 without causing 25259?

garyhgohoos: the change for 24698 was removed from xtuple and added to xwdspa

PR #2151, opened 2015-01-30
Issue #25327:add dropship argument for xwdspa

garyhgohoos:


ssshyam:
Unable to perform UAT: Since we are not able to reproduce the issue in reported version to verify the fix. Please provide more information to verify the fix.

tomdatkins:
ssshyam, I will verify the fix and close the incident. The issue is with the cost on the sales order, and it loading a "SPA COST" in the cost field under the detail tab. Gary should provide me with an updated package and I can then verify it before adding it to customers installs.
Tom

foundation-database/public/functions/itemcost.sql
```diff
@@ -7,7 +7,8 @@ CREATE OR REPLACE FUNCTION itemCost(pItemid INTEGER,
                                     pCurrid INTEGER,
                                     pEffective DATE,
                                     pAsOf DATE,
-                                    pSiteid INTEGER) RETURNS NUMERIC STABLE AS $$
+                                    pSiteid INTEGER,
+                                    pDropShip BOOLEAN) RETURNS NUMERIC STABLE AS $$
```
gilmoskowitz: This will break any caller that uses the older version of the function. Could address that by adding a default value for `pDropShip`.

garyhgohoos: i agree, but i KNOW that the only place where this is called is from salesOrderItem where the corresponding change has been made.  did not think it was worth creating an overload

gilmoskowitz: the only place in _our_ code, sure. what about extensions or other customizations?

PR #2158, opened 2015-02-03
build ci environment off xtuple server 3

shackbarth:


shackbarth:
I guess you do need those paths to npm. Assuming this is a quirk with TravisCI worth looking into in 2018.

PR #2175, opened 2015-02-09
copyQuote characteristic fix

discountscott:
copyQuote did not copy over the characteristics properly


garyhgohoos:
is this change in response to an incident?  if so, please supply the incident number so the incident can be updated with the merge.  normal commit comment is "Issue #<incident number>: <description of change>"

discountscott:
I apologize - I hadn't opened an incident for it. It's something we ran into this morning and just fixed real quick. Need me to open an issue?

garyhgohoos:
creating incident is normal procedure.  another method for tracking changes.  not necessary, merging

PR #2190, opened 2015-02-12
issue #24213 - expand compare versions function

gilmoskowitz:
Issue #24213: Revised compare version function

The new version should be callable from the Updater to compare package versions and from updater packages themselves in version number prerequisite checks. Thus
```sql
select not fetchMetricText('ServerVersion') > '4.9.0';
```
can be replaced with
```sql
select compareVersion(fetchMetricText('ServerVersion'), '4.9.0') = -1;
```
The new select is a little longer but avoids the `<` encoding issue. It also handles alpha/beta/rc/final queries and transitions between X.9.Y and X.10.0.

### DIFF
--- a/foundation-database/manifest.js
+++ b/foundation-database/manifest.js
@@ -52,6 +52,7 @@
     "public/types/wodata.sql",
     "public/types/woinvav.sql",
 
+    "public/functions/_normalizeversion.sql",
     "public/functions/acknowledgemessage.sql",
     "public/functions/actcost.sql",
     "public/functions/addrusecount.sql",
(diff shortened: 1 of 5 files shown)


davidbeauchamp:
I pulled in these changes to my 4.8.1 database and went through the tests manually, but did not receive the same error Travis is throwing. Did not receive any errors, actually...

gilmoskowitz:
Travis-CI is running Postgres 9.1, my VM is defaulting to 9.3. Turns out `regexp_matches` behaves  differently between the two (null vs. empty strings).

davidbeauchamp:
Performed Manual UAT: Issue is fixed.

PR #2220, opened 2015-02-19
Cannot add a sold item if inventoryUnit and priceUnit do not match if th...

bendiy:
...ere's no itemuomconv.

I'm not sure how we would push a `itemuomconv` in for this test item. It's not exposed to the ORM layer.


bendiy:
I think this will fix the tests for this:
https://github.com/xtuple/private-extensions/pull/567

lib/orm/source/xt/javascript/data.sql
```diff
@@ -1922,16 +1922,56 @@ select xt.install_js('XT','Data','xtuple', $$
...
+          /* Recurse into this function and try again. */
+          plv8.elog(WARNING, 'upsert duplicate key on XT.getVersion.');
+          etag = this.getVersion(orm, id);
+        }
+
+        if (!etag) {
+          /* It's possible to get no etag from the upsert due to concurrency. */
+          /* Recurse into this function and try again. */
+          plv8.elog(WARNING, 'etag upsert recursion on XT.getVersion.');
+          etag = this.getVersion(orm, id);
```
gilmoskowitz: Recursion feels wrong here in part because there's no check for a specific failure. Probably better to loop 3-5 times with early exit on success.

bendiy:
Fixed by: https://github.com/xtuple/xtuple/pull/2223

PR #2221, opened 2015-02-20
Issue #17257:correct currency conversion for allocated credits

garyhgohoos:


ssshyam:
Performed UAT: Observed that 'Invoice Had Value 0' dialog is not displayed for a foreign customer with 'Your Pay' verification Service.

PR #2222, opened 2015-02-20
fix test failures that started around the new year

gilmoskowitz:


ssshyam:
Unable to perform UAT: Please provide detailed steps.

gilmoskowitz:
@ssshyam - This is a fix specifically to the test, not the application, so no UAT is required. The bank reconciliation test was `skip`ed because the accounting periods required for creating accounting transactions did not exist. The test now creates them if necessary.

PR #2223, opened 2015-02-21
Getverion Catch ERROR duplicate key and try again. Fix item UOM test.

bendiy:


lib/orm/source/xt/javascript/data.sql
```diff
@@ -1925,13 +1925,21 @@ select xt.install_js('XT','Data','xtuple', $$
...
+
+        try {
+          plv8.execute(sql, [oid, id, etag]);
+        } catch (err) {
+          /* It's possible to get ERROR duplicate key value violates unique constraint */
+          /* "ver_ver_table_oid_ver_record_id" from the insert due to concurrency. */
+          /* Recurse into this function and try again. */
+          plv8.elog(WARNING, 'upsert duplicate key on XT.getVersion.');
+          etag = this.getVersion(orm, id);
+        }
```
gilmoskowitz: This shouldn't work. If the insert fails with a duplicate key error, inserting again should fail for the same reason unless someone removes the record in the intervening microseconds. #2220 looks like a more promising approach despite Travis-CI.

bendiy: When it recurses, it will find the existing etag in the select statement above.
https://github.com/bendiy/xtuple/blob/getverion/lib/orm/source/xt/javascript/data.sql#L1916

gilmoskowitz: But if the error is that the database server died then it'll recurse forever.

PR #2230, opened 2015-02-26
Issue #22909: Complete QuoteToInvoice process for PurchaseOrder Items

andersdd:
Removes the error that is generated in this scenario.

Creates Invoice as required and for items with the PO from SO setting enabled it also creates a Purchase Order (or adds to an existing order if already exists).

Also completed copying of Quote characteristics to the Invoice and the PO (header level only).


ssshyam:
Tried to perform UAT: Observed the following error on building the database.

building client core
building crm
building project
building sales
building billing
building purchasing
building oauth2
/home/zenqa/src/xtuple/node_modules/xtuple-dashboard-anything has no client code. Not trying to build it.
building inventory
building distribution
building manufacturing
error: undefined undefined '/home/zenqa/src/xtuple/foundation-database/foundation-database/public/functions/createpurchasetoquote.sql does not exist'
/home/zenqa/src/xtuple/foundation-database/foundation-database/public/functions/createpurchasetoquote.sql does not exist

sindhupriyaav:

foundation-database/public/functions/convertquotetoinvoice.sql
```diff
@@ -186,6 +187,31 @@ BEGIN
   FROM quhead JOIN custinfo ON (cust_id=quhead_cust_id)
   WHERE (quhead_id=pQuheadid);
 
+  -- Copy characteristics from the quhead to the invoice head
+  -- while avoiding duplicates
+  FOR _c IN
```
gilmoskowitz: I know other code uses loops but that other code should be rewritten. We should be able to do this without an explicit loop:
```
insert into charass (...)
select 'INV', _iheadid, ...
  from charass join char on charass_char_id = char_id
 where char_invoices and charass_target_type = 'QU' and charass_taget_id = pQuheadid
   and not exists(select 1 from charass where charass_target_type = 'INV' and charass_target_id = _iheadid and ...)
```
That should be a few lines shorter and a bit faster and no harder to understand.

foundation-database/public/functions/createpurchasetoquote.sql
```diff
@@ -0,0 +1,329 @@
...
+  -- Check for existing poitem for this quitem
+  SELECT poitem_id INTO _poitemid
+  FROM poitem
+  WHERE (poitem_order_id=pQuitemId)
+    AND (poitem_order_type='S');
+  IF (FOUND) THEN
+    RETURN _poitemid;
+  END IF;
+
+  SELECT *,
```
gilmoskowitz: `select *` here and below will probably break in mobilized databases because both cntct and addr tables have uuid columns.

foundation-database/public/functions/createpurchasetoquote.sql
```diff
@@ -0,0 +1,329 @@
...
+      ( _poitemid, 'U', _poheadid, _polinenumber,
+        _s.quitem_scheddate, _s.quitem_itemsite_id,
+        COALESCE(_i.itemsrc_vend_item_descrip, TEXT('')), COALESCE(_i.itemsrc_vend_uom, TEXT('')),
+        COALESCE(_i.itemsrc_invvendoruomratio, 1.00), (_s.orderqty / COALESCE(_i.itemsrc_invvendoruomratio, 1.00)),
+        _price, COALESCE(_i.itemsrc_vend_item_number, TEXT('')),
+        pItemSourceId, pQuitemId, 'Q', _s.quhead_prj_id, stdcost(_i.itemsrc_item_id),
+        COALESCE(_i.itemsrc_manuf_name, TEXT('')), COALESCE(_i.itemsrc_manuf_item_number, TEXT('')),
+        COALESCE(_i.itemsrc_manuf_item_descrip, TEXT('')), _taxtypeid,
+        COALESCE(_s.quitem_memo, TEXT('')));
+  ELSE
```
gilmoskowitz: The two sides of this if/else look identical. Either I missed something or the conditional is unnecessary.

foundation-database/public/functions/createpurchasetoquote.sql
```diff
@@ -0,0 +1,329 @@
...
+    _poheadid := _temp;
+    UPDATE pohead
+    SET pohead_dropship = pDropShip
+    WHERE (pohead_id = _poheadid);
+  ELSE
+    -- Create New
+    SELECT NEXTVAL('pohead_pohead_id_seq') INTO _poheadid;
+    SELECT fetchPoNumber() INTO _ponumber;
+
+    IF (pDropShip) THEN
```
gilmoskowitz: It would be much more maintainable to have a single `insert` here, even if it took a few more lines of code to populate a `record` variable with all of the appropriate fields to pass to the `insert`.

PR #2237, opened 2015-03-02
Cleanup any xt.foo that should not be in xt schema.

bendiy:
Fixes some of #25472


ssshyam:
Performed UAT. Please provide more information

bendiy:
@ssshyam This pull request removes some functions that were in the `xt` schema that should not be. Due to a schema search path setting in older database builds, these functions were created in the `xt` schema:
`updateCreditMemoLine()`
`insertInvoice()`
`insertInvoiceLineItem()`
`updateInvoiceLineItem()`

You will not have this problem on a new test database. It only effects some older version when they are upgraded.

PR #2239, opened 2015-03-03
update default license key

gilmoskowitz:


ssshyam:
Performed UAT: Observed that issue is not fixed for masterref databases created by using following commands.
cd xtuple
sudo ./scripts/build_app.js -d dbname -e foundation-database

gilmoskowitz:
@ssshyam - I don't think this is fixable in code. When upgrading an existing database, we should not overwrite the existing license key in case it grants broader access (such as user count) than the simple key we use for demos.

@gpazo - please update the license key in the 4.8.x masterref database backups and push those back to the shared folder for ZenQ to use. Keep the old backups in a subfolder just in case.

gpazo:
@gilmoskowitz I've updated the masterref db's for 4.8.0 and 4.8.1 on dropbox and sun.xtuple.com

PR #2246, opened 2015-03-09
Issue #24587: Invoice Margin calculated incorrectly for foreign currency...

andersdd:
... invoices

Correct cost calculation on calcinvoiceamt function to factor in alternate currencies on the invoice.

### DIFF
--- a/foundation-database/public/functions/calcinvoiceamt.sql
+++ b/foundation-database/public/functions/calcinvoiceamt.sql
@@ -1,5 +1,5 @@
 CREATE OR REPLACE FUNCTION calcInvoiceAmt(pInvcheadid INTEGER) RETURNS NUMERIC STABLE AS $$
--- Copyright (c) 1999-2014 by OpenMFG LLC, d/b/a xTuple. 
+-- Copyright (c) 1999-2015 by OpenMFG LLC, d/b/a xTuple. 
 -- See www.xtuple.com/CPAL for the full text of the software license.
 BEGIN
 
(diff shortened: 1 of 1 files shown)


sindhupriyaav:
Performed UAT: Observed that Issue is fixed in 'Unposted Invoices' screen and still exists in 'Invoice' screen. Please find the screenshot attached for more information
margincalwrongininvoicescreen-24587

andersdd:
Confirming margin issue on the invoice item screen.  This is an independent issue from that reported here and requires a change to the qt-client - not the xtuple repository - so this PR can go ahead and I will resolve the issue in a qt-client patch

foundation-database/public/functions/calcinvoiceamt.sql
```diff
@@ -30,12 +30,14 @@ BEGIN
   SELECT COALESCE(SUM(ROUND((invcitem_billed * invcitem_qty_invuomratio) *
                             (invcitem_price / COALESCE(invcitem_price_invuomratio, 1.0)), 2)), 0.0),
          COALESCE(SUM(ROUND((invcitem_billed * invcitem_qty_invuomratio) *
-                            (COALESCE(coitem_unitcost, itemCost(itemsite_id), 0.0)
+                            (currtolocal(invchead_curr_id, COALESCE(coitem_unitcost, itemCost(itemsite_id), 0.0), invchead_invcdate)
```
gilmoskowitz: The sales order might not be in the base currency so this should probably be a call to `currToCurr()`.

gilmoskowitz: Never mind... I misread the field from the `coitem` table.

PR #2249, opened 2015-03-13
incident #24642: Resolve correct release of SO number for Auto with overrd

andersdd:
Resolve correct release of SO number for Auto with override scenario

Requires qt-client PR xtuple/qt-client#637


gpazo:
merging when CI passes, +1 on code review

sindhupriyaav:

PR #2253, opened 2015-03-19
Issue #24589: Handle doc number release when Quotes use SO numbering

andersdd:
Resolve another use-case scenario around multiple open documents and their numbering


sindhupriyaav:

PR #2254, opened 2015-03-19
Issue #25527: Fix Vendor List duplicate vendors

andersdd:
But still be able to handle Vendor Addresses from the Contact and the Vendor Address area

Consider together with xtuple/private-extensions#577 which handles the xwd metasql


sindhupriyaav:

PR #2261, opened 2015-03-31
Issue #21642: Add Sale Type to Sales Reports

andersdd:
Add parameter filter and display in list

Must be merged together with xtuple/xtuple#2262
Required by xtuple/qt-client#648

### DIFF
--- a/foundation-database/public/tables/metasql/opensalesorders-detail.mql
+++ b/foundation-database/public/tables/metasql/opensalesorders-detail.mql
@@ -35,7 +35,7 @@ SELECT DISTINCT
 FROM cohead 
      JOIN custinfo ON (cohead_cust_id=cust_id) 
      JOIN custtype ON (cust_custtype_id=custtype_id)
-     JOIN saletype ON (cohead_saletype_id=saletype_id)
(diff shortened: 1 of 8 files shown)


foundation-database/public/tables/metasql/opensalesorders-detail.mql
```diff
@@ -34,6 +35,7 @@ SELECT DISTINCT
 FROM cohead 
      JOIN custinfo ON (cohead_cust_id=cust_id) 
      JOIN custtype ON (cust_custtype_id=custtype_id)
+     JOIN saletype ON (cohead_saletype_id=saletype_id)
```
gilmoskowitz: `outer join` here and elsewhere? Nothing in the database appears to require `cohead_saletype_id`.

andersdd: I thought about that.  But from what I could see all SOs had Sale Type.  That was how I found the default bug.

But I think an outer join is safer so will apply.

gilmoskowitz: Either outer join or make `cohead_salestype_id NOT NULL references saletype(saletype_id)`

PR #2262, opened 2015-03-31
Issue #25612: Add Sale Type default checkbox to set up to fix SO bug

andersdd:
SO defaulted to first alphabetically listed Sale Type when not manually set.  This could change if Sale Types were updated.

Must be merged together with xtuple/xtuple#2261
Required by xtuple/qt-client#649

### DIFF
--- a/foundation-database/public/tables/saletype.sql
+++ b/foundation-database/public/tables/saletype.sql
@@ -1,0 +1,1 @@
+SELECT xt.add_column('saletype', 'saletype_default', 'BOOLEAN', NULL::TEXT, 'public');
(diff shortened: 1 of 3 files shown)


foundation-database/public/tables/saletype.sql
```diff
@@ -0,0 +1 @@
+SELECT xt.add_column('saletype', 'saletype_default', 'BOOLEAN', NULL::TEXT, 'public');
```
gilmoskowitz: `not null default false` is much safer than leaving `saletype_default` nullable.

gpazo:
+1 looks good

PR #2272, opened 2015-04-09
handle public and xt(core) potype tables for now

gilmoskowitz:
this should allow https://github.com/xtuple/private-extensions/pull/589 to get past the basic database test

### DIFF
--- a/test/database/gotchas.js
+++ b/test/database/gotchas.js
@@ -30,10 +30,11 @@ var _ = require("underscore"),
       creds.database = databaseName;
       datasource.query(sql, creds, function (err, res) {
         assert.isNull(err);
-        assert.equal(1, res.rowCount, JSON.stringify(res.rows));
-        // TODO: we should eliminate this override as well, either by renaming and migrating
-        // the xt table, or by removing the public table if (as we suspect) it is not used.
-        assert.equal(res.rows[0].relname, "potype");
+        // TODO: fix bug 25656 then assert # rows === 0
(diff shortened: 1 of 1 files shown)


Panduk:
Please provide more information to perform UAT

gilmoskowitz:
@Panduk - This is a change to the test code, not the product. No UAT required.

test/database/gotchas.js
```diff
@@ -30,10 +30,11 @@ var _ = require("underscore"),
       creds.database = databaseName;
       datasource.query(sql, creds, function (err, res) {
         assert.isNull(err);
-        assert.equal(1, res.rowCount, JSON.stringify(res.rows));
-        // TODO: we should eliminate this override as well, either by renaming and migrating
-        // the xt table, or by removing the public table if (as we suspect) it is not used.
-        assert.equal(res.rows[0].relname, "potype");
+        // TODO: fix bug 25656 then assert # rows === 0
+        _.each(res.rows, function (elem) {
+          assert.ok(/potype(_potype_id_seq)?/.test(elem.relname),
```
jcarlin: I don't understand why `/potype(_potype_id_seq)?/`?

gilmoskowitz: That regexp matches the two expected items: `potype` is the table and `potype_potype_id_seq` is the sequence on `potype`'s primary key. We're not sure which one will come first in the query results so just loop over the results and make sure that only these two match. `potype` would match both of those...

PR #2277, opened 2015-04-15
issue #25680 - add commercialcore if necessary and make createpkgschema idempotent

gilmoskowitz:


foundation-database/public/functions/createpkgschema.sql
```diff
@@ -29,51 +27,49 @@ BEGIN
   END IF;
 
   FOR i IN ARRAY_LOWER(_tabs,1)..ARRAY_UPPER(_tabs,1) LOOP
-    _pkgtab := pname || '.pkg' || _tabs[i];
+    _pkgtab := 'pkg' || _tabs[i];
 
     IF NOT EXISTS(SELECT oid
                   FROM pg_class
                   WHERE ((relname=_pkgtab)
```
gilmoskowitz: The bug that made `createpkgschema` non-idempotent was here in the `not exists(select...)`. There is no row in `pg_class` for a table with a name literally containing the schema name and a `.`. To fix it, though, all of the queries that referred to the `_pkgtab` variable needed to be rewritten. Hopefully the new version of the function is more readable.

scripts/lib/build_database.js
```diff
@@ -50,7 +50,25 @@ var  async = require('async'),
...
+      ;
+
+      // bug 25680 - we added dependencies on a new commercialcore extension.
+      // make sure it's registered for installation if necessary
+      if (commercialPos >= 0 && commercialcorePos < 0) {
+        console.log(extensions);
+        commercialcorePath = commercialcorePath.replace(commercialRegexp, "commercialcore");
+        extensions.splice(commercialPos, 0, commercialcorePath);
+        console.log(extensions);
+      }
```
gilmoskowitz: This feels like an incredible hack. If the list of schemas includes `inventory`, make sure that `commercialcore` is in the list before it. I hoped that build_app would automatically handle prerequisites. All I see, though, is injection of a line in the resulting .sql that causes an error if the prereq is not met.

Panduk:

gilmoskowitz:
pclark says +1

PR #2279, opened 2015-04-17
ISSUE 25588

discountscott:
Stop including cancelled line items in the subtotal. Subtotal should
match how the QT client defines subtotal.


gilmoskowitz:
+1
@bendiy - you agree?

bendiy:
Looks good to me.

PR #2287, opened 2015-04-22
format date in api.bom

bixwilson:
wrap revision_date column in formatDate function


gilmoskowitz:
The CI failure probably stems from a cascading drop of the view. Check the `manifest.js` to make sure that the view definition comes before the scripts of anything that depends on the view.

PR #2292, opened 2015-04-28
tweak ownership so we can republish

gilmoskowitz:


node-datasource/lib/query/package.json
```diff
@@ -1,13 +1,13 @@
 {
   "name": "xtuple-query",
-  "version": "1.0.0",
+  "version": "1.0.3",
```
davidbeauchamp: 1.0.7 was the last one published previously, will going back in version number cause any issues?

PR #2298, opened 2015-05-01
Issue #25711: Fix incorrect Quote numbering when number not supplied

andersdd:


ssshyam:
Performed UAT and observed the following behaviour

Case1: Selecting to create only one quote at a time, the numbering sequence in api.quote is matching the quotes generated by the xTuple client

Case2: Selecting to create two quotes at a time by opening two new quote screens, the numbering sequence in api.quote is not matching the quotes generated by the xTuple client


Thanks and Regards,
Shyam

PR #2300, opened 2015-05-06
bump version

gilmoskowitz:


sindhupriyaav:
Performed UAT: Observed user is not allowed to login to the database using xTuple 4.9.0 Beta Qt Client

PR #2302, opened 2015-05-07
Move plv8 into xTuple core

bendiy:
This PR adds plv8 as a requirement to our non-mobile databases and server.


lib/orm/source/xt/functions/js_init.sql
```diff
@@ -1,10 +1,11 @@
 drop function if exists xt.js_init();
+drop function if exists xt.js_init(boolean);
 
-create or replace function xt.js_init(debug boolean DEFAULT false) returns void as $$
+create or replace function xt.js_init(debug boolean DEFAULT false, initialize boolean DEFAULT false) returns void as $wrapper$
 
 return (function () {
 
-  if (plv8.__initialized && debug !== true) {
+  if (plv8.XT && initialize !== true && debug !== true) {
```
gilmoskowitz: Why not simply `&& ! initialize && ! debug`?

bendiy: Works for me.

scripts/install_xtuple.sh
```diff
@@ -199,7 +199,8 @@ setup_postgres() {
   log "copying configs..."
 	sudo cp $PGDIR/postgresql.conf $PGDIR/postgresql.conf.default
 	sudo cat $PGDIR/postgresql.conf.default | sed "s/#listen_addresses = \S*/listen_addresses = \'*\'/" | sed "s/#custom_variable_classes = ''/custom_variable_classes = 'plv8'/" | sudo tee $PGDIR/postgresql.conf > /dev/null
-	sudo chown postgres $PGDIR/postgresql.conf
+	echo "plv8.start_proc = 'xt.js_init'" | sudo tee -a $PGDIR/postgresql.conf
+  sudo chown postgres $PGDIR/postgresql.conf
```
gilmoskowitz: @davidbeauchamp @elipup - Please make sure this gets into the admin utility.

lib/orm/source/xt/functions/show_search_path.sql
```diff
@@ -1,4 +1,7 @@
-create or replace function show_search_path() returns text as $$
+-- Cleanup old installs in xt schema.
+DROP FUNCTION IF EXISTS xt.show_search_path();
+
+create or replace function public.show_search_path() returns text as $$
```
gilmoskowitz: Why move from xt to public?

bendiy: It's actually already in public. This is another one of those functions that ended up in `xt` by mistake due to the search path setting. We were not explicit with `create or replace function show_search_path() returns text`, so this is cleaning up any old ones in `xt.show_search_path()` and insuring this is in `public`. Some other functions with this same issue were fixed in March:
https://github.com/xtuple/xtuple/commit/be5f7e3eadb4abd1c4854033fa2aef454099644d

bendiy: I found it when this test failed:
https://github.com/xtuple/xtuple/blob/4_9_x/test/database/gotchas.js#L42-L62

gilmoskowitz:
+1 but don't merge until we have clear documentation on how to install plv8 on Mac and Windows

nedlilly:
Agree wholeheartedly on the hold until Mac/Win process is good.

PR #2306, opened 2015-05-10
Issue #25392: Prevent unauthorised editing of Trial Balance table

andersdd:
Similar to gltrans table introduce trigger function to prevent Db editing of trialbal table.

Due to the nature of the trialbal table the only real check we can introduce is to check for editing closed periods.


ssshyam:

PR #2307, opened 2015-05-10
Issue #25777: Unable to convert Quotes to Sales Order

andersdd:
Looks like it was caused by the data.fetchMetric not being available in the desktop client.  Converted to a more traditional fetchmetric.


ssshyam:

PR #2310, opened 2015-05-11
Issue #25827: Fix incorrect throughput calculations on alternate Prod UoM

andersdd:
I'm going for the world record for smallest bugfix with this PR.

Consider together with private-extensions PR #608

### DIFF
--- a/foundation-database/public/functions/explodewo.sql
+++ b/foundation-database/public/functions/explodewo.sql
@@ -198,7 +198,7 @@ BEGIN
            CASE WHEN ((booitem_rnqtyper = 0) OR (booitem_invproduomratio = 0)) THEN 0
                 WHEN (NOT booitem_rnrpt) THEN 0
                 ELSE ( ( booitem_rntime /
-                         booitem_rnqtyper /
+                         booitem_rnqtyper *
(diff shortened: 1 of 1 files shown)


ssshyam:

foundation-database/public/functions/explodewo.sql
```diff
@@ -198,7 +198,7 @@ BEGIN
            CASE WHEN ((booitem_rnqtyper = 0) OR (booitem_invproduomratio = 0)) THEN 0
                 WHEN (NOT booitem_rnrpt) THEN 0
                 ELSE ( ( booitem_rntime /
-                         booitem_rnqtyper /
+                         booitem_rnqtyper *
```
gilmoskowitz: This `/` goes back at least as far as 2007 - https://github.com/xtuple/database-obsolete/blob/caaa286b3128b1d1869963e601211cab6645531e/dbscripts/functions/explodewo.sql#L188
Has this bug really been in the code this long?

andersdd: It only becomes evident when you have different Production UoM to the Inventory UoM.   So perhaps only a few customer cases??

I ran some example calculations through the entire process from Routing to WO to reports to ensure the calculations came out.

PR #2312, opened 2015-05-12
limit the searched tables to those that'll be in the search_path of a legit client

gilmoskowitz:


gpazo:
+1 looks simple enough, will merge when it passes CI

PR #2313, opened 2015-05-13
Issue #25836: Incorrect currency decimal places on Bank Rec screen

andersdd:
5 decimal places is insufficient when converting currency so switched to 6.  This was causing rounding issues for alternate currency bank accounts.

The reconcileBankAccount code formats the currency using the UOMRatio which is manually set in the Locale screen.  I am not sure this is a good idea as this ratio defaults to 5, and the user could manually set it much lower.  I think we either introduce a new decimal setup in the Locale screen or just code it differently in the Bank Rec screen.  @gilmoskowitz appreciate your thoughts here.

### DIFF
--- a/foundation-database/public/tables/metasql/bankrec-checks.mql
+++ b/foundation-database/public/tables/metasql/bankrec-checks.mql
@@ -1,7 +1,7 @@
 -- Group: bankrec
 -- Name: checks
 -- Notes: used by reconcileBankaccount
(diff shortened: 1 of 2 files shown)


gilmoskowitz:
@andersdd - UOMRatio for the exchange rate looks wrong. The C++ should use the same precision for the exchange rate as the MetaSQL, whether symbolic or hard-coded. Comments should be added to cross-reference the two so we keep them in sync.

PR #2314, opened 2015-05-13
incdt#25831

bixwilson:
Changed default remitto_country metric from ‘Country’ to ‘United
States’ to avoid validation error.


ssshyam:
Unable to perform UAT: Issue is not observed in QA databases.

gpazo:
looks good, merging

PR #2323, opened 2015-05-21
Allow manifest.js files to set a package order and grade.

bendiy:
The `manifest.js` can now optionally be formatted like this:
```
...
  "databaseScripts": [
    "xdruple/tables/pkgscript/uom.js",
    {"path": "xdruple/tables/pkgscript/uoms.js", "order": 2},
    "priv.sql"
  ]
...
```

When this is installed with `build_app.js` the `order` of `xdruple/tables/pkgscript/uoms.js` will be #2.

### DIFF
--- a/scripts/lib/util/convert_specialized.js
+++ b/scripts/lib/util/convert_specialized.js
@@ -16,14 +16,15 @@ regexp:true, undef:true, strict:true, trailing:true, white:true */
 
   var path = require('path');
 
-  var convertFromMetasql = function (content, filename, defaultSchema) {
+  var convertFromMetasql = function (content, filename, defaultSchema, script) {
     var lines = content.split("\n"),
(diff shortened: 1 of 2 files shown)


bendiy:
I'll need this in xDruple too.

scripts/lib/util/convert_specialized.js
```diff
@@ -39,9 +40,6 @@ regexp:true, undef:true, strict:true, trailing:true, white:true */
       i++;
     }
     notes = notes.substring(" Notes:".length);
-    if (notes.indexOf("must be grade 10") >= 0) {
-      grade = 10;
-    }
```
bendiy: Remove this hack and use something generic:
https://github.com/xtuple/private-extensions/pull/615/files

PR #2333, opened 2015-06-01
Add support for Characteristics and Document Associations on Item Groups

bendiy:
itemgrp-docass-char


bendiy:
Moving to commercial core. See: https://github.com/xtuple/private-extensions/pull/619

PR #2334, opened 2015-06-01
Issue #25897:include closed lines when determining s/o sched date

garyhgohoos:


bendiy:
@garyhgohoos I'm pretty sure excluding closed lines was intentional here. 
```
AND   (coitem_status NOT IN ('C', 'X'))
``` 
I remember doing this years ago because the scheduled date in List of Open Sales Orders was misleading as it didn't reflect the real (next) scheduled date on backorders when managing a list of pending order. You would bump the open line's scheduled dates to when you expected the items to be in stock, but if you included closed lines, the List of Open Sales Orders showed you the original scheduled date, not the next expected in stock date.

@davidbeauchamp What's your experience on customer expectations from the field?

gilmoskowitz:
We might need two different functions or a boolean argument to distinguish between the two cases.

garyhgohoos:
i changed to use cohead_status as conditional for including coitems.  if order closed (cohead_status=C) then all non-cancelled coitems included, otherwise, all non-cancelled, non-closed are included.

bendiy:
Sounds good.

gilmoskowitz:
@garyhgohoos - nice solution!

PR #2360, opened 2015-06-16
Issue #25944: AR Out of Balance when CM applied

andersdd:
Use the source document currency rate instead of recalculating the currency to base as of the application date.


sindhupriyaav:
Performed UAT: Observed the behavior as described in the above comment.

PR #2365, opened 2015-06-18
I think this should work, updating the submodule hash to reflect a ch…

gpazo:
…ange made in the gts-plugins directory


lib/enyo-x/lib/gts-plugins
```diff
@@ -1 +1 @@
-Subproject commit 10ed2dded0dd5b1cc3477ce617fe63d110832fb0
+Subproject commit af5908841df03bf287e41600e5530ccf758df17a
```
gilmoskowitz: I can't find this commit hash in gts-plugins history

gpazo:
interesting problem, that would require us to either A. get this commit into the main repository and grab a new hash to be used for submodule or B. fork the repo as is and add THAT submodule into our repo and update it there. 

I'd like to go with B, but we have an added layer of complexity as the submodule is hosted here: https://bitbucket.org/GlitchTechScience/gts-plugins.git

gpazo:
closing, #2370 should fix the issue

PR #2368, opened 2015-06-21
Issue #25970: PO Line Item Freight in Alt currency

andersdd:
PO Item freight was already stored in base curr.  The Post Voucher function was incorrectly converting the freight causing G/L posting errors with incorrect PO Liability entries and Incorrect freigght gain/loss entries.

### DIFF
--- a/foundation-database/public/functions/postreceipt.sql
+++ b/foundation-database/public/functions/postreceipt.sql
@@ -1,11 +1,6 @@
--- Function: postreceipt(integer, integer)
-
--- DROP FUNCTION postreceipt(integer, integer);
-
 CREATE OR REPLACE FUNCTION postreceipt(integer, integer)
-  RETURNS integer AS
-$BODY$
--- Copyright (c) 1999-2011 by OpenMFG LLC, d/b/a xTuple. 
+  RETURNS integer AS $$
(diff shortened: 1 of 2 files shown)


Panduk:

gilmoskowitz:
This fix doesn't look right. The `poitem_freight` should be in the purchase order's currency based on the [.cpp](https://github.com/xtuple/qt-client/blob/4_9_x/guiclient/purchaseOrderItem.cpp#L449) and [.ui](https://github.com/xtuple/qt-client/blob/4_9_x/guiclient/purchaseOrderItem.ui#L1647-L1651). [postReceipt](https://github.com/xtuple/xtuple/blob/164a792a3dd1e35af49a8900501c8e2d999ed578/foundation-database/public/functions/postreceipt.sql#L183) might be a cleaner target for a fix as it would make future interpretation of data in `poitem` more clear. Any bad data should clear out fairly quickly. Thoughts?

andersdd:
@gilmoskowitz Yes I think you are right.  Looks like the error started earlier in the process.  I have now fixed the postReceipt function and corrected the postVoucher function.  Unit tests show the transactions are posting to the G/L as expected.

Panduk:
UAT completed.

PR #2371, opened 2015-06-22
issue #23834 - combine all te

gilmoskowitz:
This is part three of the fix for 23834. This part uses build_app to load the current version of xtte into reference databases as we build them for a release. It depends on xtuple/xtte#15 (combines all of the xtte code in one place) and is related to xtuple/xtuple-extensions#203 (removes the now-superfluous copy).

### DIFF
--- a/scripts/lib/build_database.js
+++ b/scripts/lib/build_database.js
@@ -110,7 +110,8 @@ var  async = require('async'),
             extensionLocation: isCoreExtension ? "/core-extensions" :
               isPublicExtension ? "/xtuple-extensions" :
               isPrivateExtension ? "/private-extensions" :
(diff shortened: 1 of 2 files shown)


gpazo:
I have a couple ideas: 
add ```--``` to every passed in flag
```MAJ MIN PAT``` are kind of lost at the end of the passed in arguments, consider using the same convention ```--``` or a parsed in actual tag i.e. ```4.9.0RC``` would be interpreted ```4 , 9 , 0RC``` or equivalent?

pulled in and locally tested, :+1:

Panduk:

gilmoskowitz:
@gpazo - good idea. let's hold off on that one for a future release.

PR #2378, opened 2015-06-23
Issue #25402:round nonfractional qty to locale scale prior to ceiling

garyhgohoos:


Panduk:
Performed UAT: Issue still exists.

PR #2382, opened 2015-06-26
Issue #25656: Remove duplicated public potype table and references

andersdd:
@gilmoskowitz I would appreciate your comments on this PR.  I have removed reference to potype, and added a _drop table_ line to remove the table but not sure how this will work with the Db build process.  I've not dropped the report but we will want to do that at some stage.


foundation-database/440_schema.sql
```diff
@@ -104506,14 +104506,15 @@ ALTER TABLE public.poreject_poreject_id_seq OWNER TO admin;
...
-ALTER TABLE public.potype OWNER TO admin;
+--CREATE TABLE potype (
+--    potype_id integer NOT NULL,
+--    potype_name text,
+--    potype_descrip text
+--);
+-- ALTER TABLE public.potype OWNER TO admin;
+
+-- Issue #25656 Remove public.potype table as it is a duplicate
+   DROP TABLE IF EXISTS public.potype;
```
gilmoskowitz: Travis-CI is failing because `440_schema.sql` adds a comment to the table. Search for `potype` in this file and you'll find more potential causes for failure.

gilmoskowitz:
According to github's code search there are 7 files that refer to `potype`:

- enyo-client/extensions/source/purchasing/database/source/xt/tables/potype.sql
- enyo-client/extensions/source/purchasing/database/source/xt/tables/poheadext.sql 
- enyo-client/extensions/source/purchasing/database/source/xt/tables/potypewf.sql 
- test/database/gotchas.js
- enyo-client/extensions/source/purchasing/database/source/manifest.js 
- enyo-client/extensions/source/purchasing/database/orm/models/purchase_order.json
- database/source/xt/trigger_functions/createwf_after_insert.sql

The 4 files changed in this PR aren't even in the list?!

andersdd:
@gilmoskowitz The files you raise all point to xt.potype which is the mobile po type that is actually being used (for workflow).  I have removed reference to public.potype which from what I could see from searches was either a mistake or just not used.

gilmoskowitz:
I missed that.... However, `xt.potype` doesn't exist in the xtuple repository so these are forward references to a column or table that won't exist unless and until the private extensions are loaded. Sounds like SQL errors waiting to happen. How could this have worked at all over the past couple of months?

andersdd:
@gilmoskowitz  xt.potype is created in enyo-client/extensions/source/purchasing/database/source/xt/tables/potype.sql  not private-extensions which is why the remaining logic works.

gilmoskowitz:
@andersdd - it's also created in https://github.com/xtuple/private-extensions/blob/4_10_x/source/commercialcore/foundation-database/xtcore/tables/potype.sql  but at least this time it appears to be consistent with the version in enyo-client/extensions.

andersdd:
@gilmoskowitz I would suggest the commercialcore version of xt.potype is now redundant and can probably be removed.  The potype table is created as part of xtuple/extensions/purchasing so is core logic.

gilmoskowitz:
@andersdd - fair 'nuf. I'll merge this PR. Please submit another to remove the superfluous copy in private-extensions (just use the same bug number). Thanks.

andersdd:
@gilmoskowitz I think the commercialcore version of potype is for non-mobilised databases so we may still need this code.  It does not cause any real issues by being there.

PR #2385, opened 2015-06-27
issue #26021, issue #26064, issue #15457 -- issue with backslash escaping …

davidbeauchamp:
…in PostgreSQL. Requires commits in qt-client and openrpt as well


davidbeauchamp:
Pulling in these changes means the windows build needs to be Qt 4.8.6! The way we were dealing with escaping only worked with Qt < 4.8.6.. Since they fixed the PostgreSQL driver we don't need to hack around it.

PR #2387, opened 2015-06-29
Issue #26085:concatenate order number and line number to populate glt…

garyhgohoos:
…rans docnumber


Panduk:

PR #2400, opened 2015-07-08
issue #26132 - preserve existing source records, add f-keys

gilmoskowitz:
and ensure upgrades can create 'em


foundation-database/public/functions/createdoctype.sql
```diff
@@ -12,35 +12,36 @@ create or replace function createDoctype(pDocAssNum integer, pType text, pDocAss
...
          pDesc,   pJoin,   pWidget,
          pParam,  pUi,     pPriv)
       returning source_id into _id;
-  elsif _justAddedCols then
+  else
     update source
-       set source_docass_num   = pDocAssNum,
+       set source_docass_num   = coalesce(pDocAssNum, 0),
            source_docass       = pDocAss,
            source_charass      = pCharass,
```
bendiy: Include `source_descrip = pFull,` ?

bendiy: Would we ever need to change `source_module = pModule`?

foundation-database/public/tables/cmnttypesource.sql
```diff
@@ -0,0 +1,15 @@
+select xt.create_table('cmnttypesource', 'public');
+select xt.add_column('cmnttypesource', 'cmnttypesource_id', 'serial', 'primary key', 'public');
+select xt.add_column('cmnttypesource', 'cmnttypesource_cmnttype_id', 'integer', NULL, 'public');
+select xt.add_column('cmnttypesource', 'cmnttypesource_source_id', 'integer', NULL, 'public');
```
bendiy: Can you add comments to the column? The 6th param of `add_column` is a comment:
https://github.com/gilmoskowitz/xtuple/blob/i26132_cmnttypesource/lib/orm/source/xt/functions/add_column.sql#L3

gilmoskowitz: Comments on the columns wouldn't be particularly helpful since they're already foreign keys and the table is narrow. Instead I expanded the comment on the table itself to clarify its purpose.

PR #2402, opened 2015-07-09
2198

gpazo:


gpazo:
doh, 4_9_x forward merge got wrapped up in my changes...

PR #2403, opened 2015-07-09
validate row prior to creating new line item, model changes to valida…

gpazo:
…te by Julian

### DIFF
--- a/lib/enyo-x/source/views/grid_box.js
+++ b/lib/enyo-x/source/views/grid_box.js
@@ -191,7 +191,7 @@ trailing:true, white:true, strict: false*/
     keyUp: function (inSender, inEvent) {
       var model = this.value;
       // Don't navigate line items if the current model doesn't have all it's req. attrs.
-      if (!model.validate(model.attributes)) {
+      if (model && !model.validate(model.attributes)) {
         if (inEvent.keyCode === XV.KEY_DOWN) {
           this.doMoveDown(inEvent);
           return true;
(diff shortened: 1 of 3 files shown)


lib/enyo-x/source/views/grid_box.js
```diff
@@ -189,14 +189,20 @@ trailing:true, white:true, strict: false*/
...
+          this.doMoveDown(inEvent);
+          return true;
+        } else if (inEvent.keyCode === XV.KEY_UP) {
+          this.doMoveUp(inEvent);
+          return true;
+        } else if (inEvent.keyCode === XV.KEY_ENTER) {
+          this.doEnterOut(inEvent);
+          return true;
+        }
+        // Exit the event chain if req. attrs.
```
jcarlin: Oops l205, l206 should be moved to below/outside the `if`.

Panduk:

lib/enyo-x/source/views/grid_box.js
```diff
@@ -189,16 +189,22 @@ trailing:true, white:true, strict: false*/
...
-        return true;
-      } else if (inEvent.keyCode === XV.KEY_UP) {
-        this.doMoveUp(inEvent);
-        return true;
-      } else if (inEvent.keyCode === XV.KEY_ENTER) {
-        this.doEnterOut(inEvent);
-        return true;
+      var model = this.value;
+      // Don't navigate line items if the current model doesn't have all it's req. attrs.
+      if (!model.validate(model.attributes)) {
```
jcarlin: `if (!model.validate(model.attributes)) {`
change to
`if (model && !model.validate(model.attributes)) {`

lib/enyo-x/source/views/grid_box.js
jcarlin: This comment should be moved to above the `return true;` on l207

PR #2404, opened 2015-07-10
Issue #25898: SQL Error on PO Quick Entry where no Item Source exists

andersdd:
Scenario where no item price is entered AND no Item Source exists.


Panduk:

PR #2405, opened 2015-07-10
Issue #20129: Currency Exchange Rate interface

andersdd:
@gilmoskowitz This meets the requirement as specified in the last comment on the issue.

Not sure if we want to do more??  I know @davidbeauchamp has done a funky exchange rate import script using some FX webservice.  I think it is currently used in dogfood.

I have left the issue marked as open (unresolved) at the moment until we decide if anything else is required.

### DIFF
--- a/foundation-database/api/views/currencyrate.sql
+++ b/foundation-database/api/views/currencyrate.sql
@@ -1,0 +1,32 @@
+SELECT dropIfExists('VIEW', 'currencyrate', 'api');
+CREATE OR REPLACE VIEW api.currencyrate AS
+  SELECT curr_abbr AS currency,
+     CASE WHEN (fetchmetricvalue('CurrencyExchangeSense') = 1) THEN
+               round((1.0 / curr_rate),5)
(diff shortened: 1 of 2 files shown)


davidbeauchamp:
I wrote an integration to https://openexchangerates.org/ from inside the Qt client, which will fetch exchange rates for a given time period for all currencies defined in the system. Open Exchange Rates is free for 1000 API hits a month, as long as your base currency is USD. The fee schedule is minimal if you wanted a different base currency.

gilmoskowitz:
@andersdd - thanks.

@davidbeauchamp - did you commit your openexchangerates integration somewhere? we've been talking a long time about something like and it would be nice to publish it somewhere/somehow.

andersdd:
@gilmoskowitz  I think David sent me a copy of the package a while ago.  I was questioning [a] whether we want to take exchange rates further (in core or as an extension package) and [b] whether @davidbeauchamp would give permissions to use the package.

I also have a web service I have been using for about 5 years using Pentaho integration, I can port that to either core or a js package if necessary.

PR #2422, opened 2015-07-24
CRITIAL!!! Fix getsoscheddate() performance regression. Effects all SO lists.

bendiy:
getSoSchedDate(INTEGER) was selecting ALL cohead orders to get the status of one order.


bendiy:
List of Open Sales Orders was taking 5-10 minutes to load on large databases.

gilmoskowitz:
@bendiy - This set of changes isn't quite right. @bartmaj and I were looking at this function last week. Let's talk more about this off-line.

PR #2430, opened 2015-08-03
issue #26212 - fix remaining uses of char_* bool columns to use charuse

gilmoskowitz:


Panduk:

PR #2431, opened 2015-08-04
Remove npm from our package.json. It should never have been there.

bendiy:
I'm not sure how disruptive this will be. To cleanup installs, you will need to run:
```
sudo rm -r node_modules/
sudo n 0.10.40
sudo npm install npm -g
npm install
```

That should install Node.js version `0.10.40` and npm version `2.13.3`. The version we had been using, `1.4.x` is very out dated and full of bugs. This PR's `npm-shrinkwrap.json` was made using the latest versions.


bendiy:
Finally got this change to pass, but let's wait to merge until I can do some tests with private-ext too.

bendiy:
@gilmoskowitz This should be ready to merge now. I tested on pri-ext locally and all passed.

gilmoskowitz:
@bendiy - Before this can be merged, the dependencies on specific node and npm versions need to be codified. As absolute minimum, install_xtuple.sh must be updated to install the newly correct versions.

bendiy:
@gilmoskowitz: `install_xtuple.sh` has been updated. Tests were passing on the older version of node and npm as well, so this is more of an optional upgrade to node `0.10.40` and npm `2.13.4`.

PR #2433, opened 2015-08-05
Issue 26223: Extend Item pricing to Shipping Zone and Sale Type

andersdd:
Add database components to support Sale Price assignment by Shipping Zone or Sale Type


foundation-database/public/functions/itemcharprice.sql
```diff
@@ -44,7 +44,7 @@ DECLARE
 BEGIN
   RETURN itemCharPrice(pItemid, pCharid, pCharValue, pCustid, pShiptoid, pQty, pCurrid, CURRENT_DATE);
 END;
-$$ LANGUAGE plpgsql;
+$$ LANGUAGE 'plpgsql';
```
gilmoskowitz: Better to leave off the quotes around the language name. Postgres **9.1** docs say they're allowed for backward compatibility.

foundation-database/public/tables/ipsass.sql
```diff
@@ -0,0 +1,6 @@
+select xt.add_column('ipsass', 'ipsass_shipzone_id', 'integer', 'DEFAULT -1', 'public', null);
+select xt.add_column('ipsass', 'ipsass_saletype_id', 'integer', 'DEFAULT -1', 'public', null);
```
gilmoskowitz: These should both be foreign keys to the shipzone and saletype tables, therefore NULLable. This will ripple through the PR. The -1 convention causes trouble.

PR #2436, opened 2015-08-05
update install_xtuple.sh to install openrpt by default, following fun…

gpazo:
…ction convention already present


Panduk:
Performed UAT: Observed that openrpt installed successfully.

davidbeauchamp:
this will need xvfb if openrpt is expected to run

PR #2439, opened 2015-08-07
Issue #23152: Add privilege allowing to edit/create shared filters

bartmaj:
Core app changes at: xtuple/qt-client#801


foundation-database/public/functions/grantpriv.sql
```diff
@@ -62,3 +62,36 @@ BEGIN
...
+ 			 SELECT _p.pUsername, priv_id
+   			 FROM priv
+  				 WHERE (priv_name=pPrivname);
+
+ 			 NOTIFY "usrprivUpdated";
+	END LOOP;	
+RETURN TRUE;
+
+END;
+$$ LANGUAGE plpgsql;
```
gilmoskowitz: This is a new function name and so belongs in its own file. In addition almost all of the logic can be eliminated by simply calling `SELECT grantPriv(usr_username, pPrivname) FROM usr;` If we're going to give up if any one grant fails, use the `BOOL_AND` aggregate function. If the grant should do as much work as possible, then loop through and accumulate the return value.

PR #2443, opened 2015-08-11
Issue #26247:correctly recall TO shipments conditionally by status of…

garyhgohoos:
… receipt


ssshyam:
Unable to confirm the fixed behavior as issue is not observed on QA databases.

PR #2444, opened 2015-08-11
issue #26313 - source_docass_num must match c++ enum, crmacct and child windows should show each others docass

gilmoskowitz:
Note that this is not a complete fix for 26313. There are also some changes required in the desktop client to complete that - https://github.com/xtuple/qt-client/pull/813 - but they can be merged independently.


Panduk:
Performed UAT:Issue is fixed

andersdd:
+1 code review. I've not unit tested but the logic looks sound

PR #2446, opened 2015-08-13
Only select the current revision of bomhead as multiple revisions wil…

bendiy:
…l return more than one row for the subselect.

PR in xtdesktop here: https://github.com/xtuple/xtdesktop/pull/38


gilmoskowitz:
Extensions and their components should not be maintained in the reference database source files.

PR #2450, opened 2015-08-14
Remove query and congruence. The indirection is toxic.

bendiy:
Replace the query library with simple version that also works in plv8.

He's the toughest S.O.B. to ever wear a tutu!


lib/orm/source/xm/javascript/model.sql
```diff
@@ -376,129 +376,230 @@ select xt.install_js('XM','Model','xtuple', $$
...
+          validateColumn = function (column) {
+            // Make sure the column name is a valid PostgreSQL string.
+            // Make sure the column is in the schema for this recordType.
+            var valid = false;
+
+            if (validColNamePattern.test(column)) {
+              for (var c = 0; c < schema.columns.length; c++) {
+                // TODO: Only checking the recordType parameter columns, not any
+                // toOne/toMany child columns. e.g. "[address.city]"
+                if (schema.columns[c].name === column.split(".")[0]) {
```
gilmoskowitz: `column.split()` outside the loop.

Where is `schema.columns` populated? `validateColumn()` looks like it'll get called frequently so sorting that array on creation then using a binary search here seems warranted.

lib/orm/source/xm/javascript/model.sql
```diff
@@ -376,129 +376,230 @@ select xt.install_js('XM','Model','xtuple', $$
...
+          validateValue = function (value) {
+            // The value should be a string, boolean or numeric.
+            switch (typeof value) {
+              case "undefined":
+              case "symbol":
+              case "object":
+              case "function":
+                throw new Error();
+              default:
+                return value;
```
gilmoskowitz: Whitelisting seems safer than blacklisting.

node-datasource/routes/restRouter.js
```diff
@@ -2,22 +2,16 @@
...
+  _ = require("underscore"),
+  restQueryFormat = require("../lib/rest-query-format");
 
 /**
  * @class RestRouter
  * @singleton
  */
 module.exports = (function () {
-  'use strict';
+  "use strict";
```
gilmoskowitz: Why change quote character? It generates a lot of churn.

bendiy: @gilmoskowitz restRouter.js had a mix of `'` and `"`. I'm trying to be consistent throughout the file with these changes.

PR #2453, opened 2015-08-17
Issue #26224:add qty reserved

garyhgohoos:


gilmoskowitz:
see xtuple/qt-client#815 and xtuple/private-extensions#682

PR #2455, opened 2015-08-18
Issue #26283:display actual scrap instead of expected scrap %

garyhgohoos:


gilmoskowitz:
goes with xtuple/qt-client/#817

PR #2456, opened 2015-08-18
Issue #26330:add quote parent type

garyhgohoos:


gilmoskowitz:
see xtuple/qt-client#818

PR #2472, opened 2015-08-24
Issue #25686: Resolve meaningless error msg when posting GL Series

andersdd:
Return meaningful message when posting to non-existant period


Panduk:
Performed UAT:Issue is fixed

PR #2474, opened 2015-08-25
Issue #14182: Post InvHist record when updating Standard Cost

andersdd:
Also resolves issue #26236

Posts a inventory value only change (nil qty. change) to inventory history so that the Inventory ledger and the General Ledger reconcile.


Panduk:
Performed UAT: Observed that 'SC' type transactions in Inventory History screen are according to the entries made in GL screen on adjusting Standard Cost of an item.Please observe that Incident # 26236 is an invalid number, please provide the correct incident number

gilmoskowitz:
Incident 26236 is a support ticket, not a bug report or feature request, and so is not generally visible. It is appropriate:-)

PR #2475, opened 2015-08-25
Issue #26377:account for kit components that have been issued

garyhgohoos:


Panduk:
Performed UAT:Issue is fixed

foundation-database/public/functions/qtyatshipping.sql
```diff
@@ -1,10 +1,26 @@
...
-  RETURN qtyAtShipping('SO', $1);
+  -- account for kits by including kit components
+  SELECT coitem_cohead_id, coitem_linenumber INTO _coheadid, _linenumber
+  FROM coitem
+  WHERE (coitem_id=plineitemid);
+
+  SELECT COALESCE(SUM(qtyAtShipping('SO', coitem_id)), 0.0) INTO _qty
+  FROM coitem
+  WHERE (coitem_cohead_id=_coheadid)
+    AND (coitem_linenumber=_linenumber);
```
gilmoskowitz: This doesn't look right. It addresses the bug as reported but seems as though it would break a lot of places that use the result for something other than a `> 0` check.

garyhgohoos: i agree.  please suggest a solution.  i can think of a couple; 1. additional comments that explain that this overload of the function includes kit components and a different overload should be used if kit components should not be included  2. create a new overload specifically for including kit components

gilmoskowitz: I'd suggest writing a different function, one designed to answer the question "does X have anything at shipping?" Use the `qtyAtShipping()` functions for exact numbers for line items and this new function or set of functions to address "safety" questions like 26337. The new function should be faster since it can be written as `SELECT EXISTS(SELECT 1 FROM coitem WHERE ...)...` and can probably be written as a SQL function, rather than plpgsql (SQL functions get inlined, iirc).

PR #2477, opened 2015-08-27
Issue #18310:create wo according to planning parameters

garyhgohoos:


Panduk:
Performed UAT:Issue is fixed

PR #2478, opened 2015-08-27
Issue #26421: Fix qtyNetable function which was returning NULL 

andersdd:
Inventory solely located in a non-netable location resulted in this function returning NULL and preventing MPS schedule from working.

@garyhgohoos would you mind checking my logic please?


garyhgohoos:
looks good

PR #2489, opened 2015-09-02
Cleanup various cruft

bendiy:


node-datasource/scripts/node
```diff
@@ -1,60 +0,0 @@
...
-  restart|force-reload)
-    $0 stop
-    $0 start
-    ;;
-  *)
-    log_action_msg "Usage: /etc/init.d/$NAME {start|stop|reload|restart|force-reload|syntax|status}"
-    ;;
-esac
-
-exit 0
```
gilmoskowitz: This file really isn't used anywhere?!?

node-datasource/xt/foundation/foundation.js
```diff
@@ -370,18 +358,6 @@ X = {};
...
-      X.writeFile(X.pidFile, X.pid);
-      X.addCleanupTask(X.cleanupPidFile);
-    },
-
-    cleanupPidFile: function () {
-      if (!X.pidFile) return;
-      X.log("Removing pid file.");
-      X.removeFile(X.pidFile);
-    },
-
```
gilmoskowitz: Who manages the pid files now?

bendiy: I tested this and it didn't seem to be working anymore anyways. At one point, you couldn't start the datasource if it was already running. The way it checked that was the pid files. That functionality has long since been broken or working another way. Currently you'll get an error that you can't start the server on it's config port, if another process is already running.

node-datasource/scripts/node
bendiy: We use init services now. `/etc/init` in admin-utli and xtuple-server.

PR #2495, opened 2015-09-13
Issue #20135: VAT Taxation enhancements

andersdd:
Initial commit.  Still needs some work following the PO Freight taxation changes (to come), but this piece stands on its own.


foundation-database/public/functions/postcheck.sql
```diff
@@ -126,12 +130,46 @@ BEGIN
...
+        IF (COALESCE(_credit_glaccnt, -1) < 0) THEN
+          RETURN -13;
+        END IF;
+
+        INSERT INTO checkheadtax (taxhist_basis,taxhist_percent,taxhist_amount,taxhist_docdate, taxhist_tax_id, taxhist_tax, 
+                             taxhist_taxtype_id, taxhist_parent_id, taxhist_journalnumber ) 
+          VALUES (0, 0, 0, current_date, _tax.taxdetail_tax_id, (_tax.taxdetail_tax * -1), getadjustmenttaxtypeid(), 
+              pCheckid, _journalNumber);
+              
+      END LOOP;
```
gilmoskowitz: Can this be written successfully as
```
INSERT INTO checkheadtax (...)
SELECT ...
FROM calculatetaxdetail(...);
```
? That approach would be a little faster, more compact, and more in keeping with the spirit of SQL.

That does raise the question, though, of how to handle the `_credit_glaccnt` error check. If the `insert ... select` doesn't work out, line 149 should be changed to `raise exception` (the way of the future).

foundation-database/public/tables/metasql/taxReturn-detail.mql
```diff
@@ -0,0 +1,417 @@
...
+  <? value("sales") ?> AS source,
+  CASE
+    WHEN (cohist_doctype='I') THEN
+      <? value("invoice") ?>
+    WHEN (cohist_doctype='C') THEN
+      <? value("creditmemo") ?> 
+    WHEN (cohist_doctype='D') THEN
+      <? value("debitmemo") ?>
+    ELSE
+      <? value("other") ?>
```
gilmoskowitz: Probably better to just let `cohist_doctype` shine through. That way the "other" case is debuggable for us and gives a hint to the poor user.

PR #2508, opened 2015-09-25
Fix performance of docinfo view.

bendiy:
This takes a single Item Master open request for Document Associations from ~5-6 seconds down to ~250ms.
https://github.com/xtuple/qt-client/blob/b02baada16a55f215bef45f392836060782ec42b/widgets/documents.cpp#L449

This will need forward merged to `4_10_x`.

Still need to grep for all queries to `docinfo` and make sure `WHERE` clauses are used to filter by `source_id` AND `source_type`. A full `SELECT * FROM docinfo` will probably die.


bendiy:
A quick search of the repos and github doesn't show any other uses. Mobile gets document through this view:
https://github.com/xtuple/xtuple/blob/4_10_x/database/source/xt/views/doc.sql

bendiy:
`SELECT * FROM docinfo;`
Still running after 2,400,000 ms. I don't know of any way to protect against that. This view can now be used ONLY for single object document associations.

gilmoskowitz:
Selecting _all_ document associations is a huge task so let's not worry about the performance of selecting without a where clause.

PR #2514, opened 2015-10-02
Issue #26391: Resolve orphaned Documents/Characteristics/Comments 

andersdd:
When cancelling or deleting a S/O documents/characteristics/comments remained in the Db as orphans.

This PR also resolves the issue when you delete a Document with an assigned File or URL - that file/url was also left orphaned in the database.  There was a question over whether we wanted the file to remain in the Db orphaned if it was something important.

### DIFF
--- a/foundation-database/manifest.js
+++ b/foundation-database/manifest.js
@@ -1067,6 +1067,7 @@
     "public/tables/cmitem.sql",
     "public/tables/cmnttypesource.sql",
     "public/tables/company.sql",
+    "public/tables/docass.sql",
     "public/tables/ipsass.sql",
     "public/tables/itemsite.sql",
     "public/tables/itemuomconv.sql",
--- a/foundation-database/public/functions/deleteso.sql
+++ b/foundation-database/public/functions/deleteso.sql
@@ -22,6 +22,17 @@ DECLARE
   _poStatus     INTEGER := 0;
 
 BEGIN
+-- These elements can be created before cohead record exists
+  DELETE FROM docass WHERE docass_source_id = pSoheadid AND docass_source_type = 'S';
+  DELETE FROM docass WHERE docass_target_id = pSoheadid AND docass_target_type = 'S';
+
+  DELETE FROM comment
+  WHERE ( (comment_source='S')
+   AND (comment_source_id=pSoheadid) );
+
+  DELETE FROM charass
+  WHERE (charass_target_type='SO')
+    AND (charass_target_id=pSoheadid);
 
 -- Get cohead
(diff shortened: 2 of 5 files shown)


gilmoskowitz:
@daveanderson: Instead of deleting files because they are unreachable, we have to build an interface to browse the existing files. Regarding the point raised offline by @bendiy about file access, let’s extend the all/personal priv set to files unless someone has a better idea.

bendiy:
@gilmoskowitz @andersdd 
What should personal files access query/filter on? Once a file is orphaned, there's no way to derive a relationship for personal ownership. Even if it's not orphaned the query to derive personal ownership could get quite convoluted. This sounds like a problem the Shared User Access feature could solve, but I don't want to port that to core. Row Level Security would be a good solution when it's available to use in a few years.

I think "all file" access should be granted sparingly. Imaging you scan employee's drivers license and Social Security card and attach then to the Employee Record as File Document Associations. Someone with access to view all files (orphaned or not) would be able to see these files.

gilmoskowitz:
@bendiy 
Good points. However, I'm not suggesting that we create a new problem. The issues you describe already pertain to images, for example. If we add fields to the `file` table for creator, create timestamp, and modified timestamp then we have the basis for the scheme we already use elsewhere. We can discuss the merits of Shared User Access and Row Level Security offline. In the meantime, there are other reasons to add auditing fields to the `file` table _and_ we can use them for a first-pass privilege implementation.

Regardless, we need at least one way to browse the `file` table so users can associate a file with multiple documents, even if the they can't read the files. Deleting orphaned `file` records because we can't reach them is the wrong solution.

See incidents 25092 and 25607 for starters.

andersdd:
OK I will remove the delete from the trigger for the moment and add in the creator and timestamp instead pending some thoughts on how we can attach files to mulitple documents (which should also resolve the orphan problem) and file permissions.

I wrote something for xtConnect (a while ago on a customer site), that restricted emails to those associated with the user, unless you had a priv. to see all emails.  It was rudimentary, but was in the same vein as our discussions here.  Might be a starting point for permissions??

andersdd:
@bendiy @gilmoskowitz  New commit removes FILE deletion on deleting the docass record.  I also added in the username and timestamp to document record.

HOWEVER, in further testing I found pressing the DETACH button deletes the file from the system (after a warning) so our discusison above is a bit of a moot point.   The scenario I was originally testing was the addition of documents to a SO then not saving the SO - this (used to) leave the document association linked to a non-existing SO, so my trigger now removes that association.  The problem now, is the file remains orphaned in the system - which is what we now need to address.

PR #2519, opened 2015-10-09
Issue #25565: Purchase Order / Voucher Freight taxation

andersdd:
Extend freight taxation from PO through to Voucher.  Also introduce header level freight.

Builds on VAT tax project so that should be checked/merged first

requires xtuple/xtuple#2495


foundation-database/public/functions/postvoucher.sql
```diff
@@ -3,19 +3,6 @@ CREATE OR REPLACE FUNCTION postVoucher(INTEGER, BOOLEAN) RETURNS INTEGER AS $$
...
-  RETURN postVoucher(pVoheadid, fetchJournalNumber('AP-VO'), pPostCosts);
-END;
-$$ LANGUAGE 'plpgsql';
-
-
-CREATE OR REPLACE FUNCTION postVoucher(INTEGER, INTEGER, BOOLEAN) RETURNS INTEGER AS $$
--- Copyright (c) 1999-2015 by OpenMFG LLC, d/b/a xTuple.
--- See www.xtuple.com/CPAL for the full text of the software license.
-DECLARE
-  pVoheadid ALIAS FOR $1;
```
gilmoskowitz: This deletion doesn't look right. It eliminates the three-parameter version of the function and creates a two-parameter version that tries to `alias for $3`.

gilmoskowitz:
Also, CI is failing. It feels like a mismatched dollar-quote or missing semicolon but I don't see any.

andersdd:
@gilmoskowitz  Looks like CI fail caused by entry in manifest file where the corresponding file was missing.  Perhaps clearer errors in this scenario required?

PR #2523, opened 2015-10-15
try to make manifest processing errors easier to find in ci output

gilmoskowitz:
see https://github.com/xtuple/xtuple/pull/2519#issuecomment-147242434 for motivation


gilmoskowitz:
@andersdd - Could you please review this? Would this have helped find [the p/o & voucher freight taxation ci failure](https://github.com/xtuple/xtuple/pull/2519#issuecomment-147242434) quicker?

andersdd:
@gilmoskowitz  Interesting error output, but I tested this and it gave me the exact cause of the error in the manifest file.  So immediate identification of error cause.

+1 revew

PR #2530, opened 2015-10-27
Issue #26757: PO Return Valuation

andersdd:
PO Return for inventory controlled items was valuing at standard cost not Purchase Order Item price.  This means Credit Memos created as a result do not match the PO and Voucher pricing.

### DIFF
--- a/foundation-database/manifest.js
+++ b/foundation-database/manifest.js
@@ -1412,6 +1412,7 @@
     "public/tables/metasql/taxAuthorities-detail.mql",
     "public/tables/metasql/taxBreakdown-detail.mql",
     "public/tables/metasql/taxHistory-detail.mql",
+    "public/tables/metasql/taxReturn-detail.mql",
     "public/tables/metasql/timePhasedAvailability-detail.mql",
     "public/tables/metasql/timePhasedBookings-detail.mql",
     "public/tables/metasql/timePhasedSales-detail.mql",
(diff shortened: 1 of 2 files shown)


foundation-database/manifest.js
```diff
@@ -1412,6 +1412,7 @@
     "public/tables/metasql/taxAuthorities-detail.mql",
     "public/tables/metasql/taxBreakdown-detail.mql",
     "public/tables/metasql/taxHistory-detail.mql",
+    "public/tables/metasql/taxReturn-detail.mql",
```
davidbeauchamp: this commit sneak in?

andersdd: Yes but it is already in 4.10 from another PR so I left it.  

Dude!! I am in China working on this from the back of a van  :-)

Panduk:
Performed UAT: Observed the following behavior.
1. Observed the following transactions in G/L on posting a PO return where the item (say 'TBOX1') has purchase price( say '0.25$' for Item Source 'TPARTS') and standard cost (say '1.00$')
###please refer the screenshot###
gl transactions

andersdd:
@panduk  I made no changes to the G/L postings.  These should still be posting based on standard cost.  If you look at the Uninvoiced Receipts and Returns screen, you should see the PO Return lines that you posted and these should be using the PO Item unit cost.  PO item unit costs may or may not be the same as the Item Source.  If you create a Credit Memo for the Return it should reflect the same unit price as the PO Item.

Panduk:
@andersdd Thanks for the information, Performed UAT: Issue is fixed

PR #2538, opened 2015-11-05
Issue #26848: Add missing RR transaction to Usage Statistics report

andersdd:
Altered underlying function so may well fix other things that use this function.


rkannadevara:
Performed UAT: Observed that there is no 'Return Receipts' transaction type filter in 'Usage Statistics' and 'Inventory History' screen. Please refer the attached screenshot: 

rr transaction type

gilmoskowitz:
Re: UAT comments - The change ensures that a particular subset of receipts gets included in the total reported for all receipts.

PR #2539, opened 2015-11-05
Issue 26849: Extract Item Groups populate query to MetaSQL, provide checkbox to show only top level item groups

davidbeauchamp:


davidbeauchamp:
Requires https://github.com/xtuple/qt-client/pull/928

andersdd:
+1 code review

PR #2565, opened 2015-12-07
Issue #25441: api.itemsite Update performance

andersdd:
This resulted in a 260x performance improvement in testing

### DIFF
--- a/foundation-database/api/views/itemsite.sql
+++ b/foundation-database/api/views/itemsite.sql
@@ -337,7 +337,8 @@ UPDATE itemsite SET
          ELSE
            'M'
        END
-   WHERE (itemsite_id=getItemSiteId(OLD.site,OLD.item_number));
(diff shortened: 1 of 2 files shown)


davidbeauchamp:
It looks like [getitemsiteid](https://github.com/xtuple/xtuple/blob/4_10_x/foundation-database/public/functions/getitemsiteid.sql) can be improved little bit by removing the SELECT * from [here](https://github.com/xtuple/xtuple/blob/4_10_x/foundation-database/public/functions/getitemsiteid.sql#L33). It looks like the function is only using itemsite_id, item_active, item_sold, itemsite_active and itemsite_sold. Worth doing that as part of this PR, too?

andersdd:
The getitemsiteid function was not the cause of the performance issues.  It was the use of a function in the WHERE clause that killed it.  Postgres is running the function for each row in the itemsite table.

I will check where else the function is used, and check out your suggestions, but we cany use this or any function in the where clause of the api view.

PR #2575, opened 2015-12-14
Issue #26375: Tidy (clarify) up report labels

andersdd:


Panduk:
Performed UAT: Issue is fixed when a single customer is selected, also observed that on selecting to print for all customers 'Totals' is displayed instead of 'Totals for all customers'
####please refer the attached screenshot####
pull on aging

andersdd:
The label "Total" is more correct.  If the report is filtered by customer type or something then this total is not for ALL customers just some of them.  I figured just adding Total was less ambiguous

PR #2581, opened 2015-12-31
Issue #26010: Purchase Order crashes xTuple application

andersdd:
I suspect this is a combination of the number of line items and the length of the PO Note field.  Something to do with one report section crossing over another (??).   I have added an arbitrary length restriction to the PO Report which resolves the issue in testing, but may not resolve under some scenarios.


sindhupriyaav:
Performed UAT: Issue is fixed. Also Observed that PO Notes is being clipped in the Print Report.

gilmoskowitz:
I've been able to cause report failures with extremely long comments for purchase orders and invoices. I'm going to close this P/R without merging it since it doesn't actually fix the bug, which is somewhere in OpenRPT.

PR #2582, opened 2015-12-31
Issue #25978: fix GL Transactions report

andersdd:


foundation-database/public/tables/report/GLTransactions.xml
```diff
@@ -97,7 +97,9 @@ SELECT gltrans_id,
        ELSE 0 END AS running,
        formatBoolYN(gltrans_posted) AS f_posted,
        formatBoolYN(gltrans_deleted) AS f_deleted,
-       gltrans_username AS f_username,
+       &lt;? if exists("showUsernames") ?>
+         gltrans_username AS f_username,
+       &lt;? endif ?>
```
gilmoskowitz: This is part of a `UNION`. The query will fail if the first `SELECT` in the union is enabled by MetaSQL parameters. It's probably better to switch between `NULL::text` and `gltrans_username` here if `showUsernames` exists, as is done elsewhere in this report.

rkannadevara:

PR #2585, opened 2016-01-07
Issue #26687:clean up reservation uom conversion

garyhgohoos:


rkannadevara:
Performed UAT: Observed that reserved qty is being displayed as '4' instead of '16' as expected. Also observed that reserve UOM is expressed in Qty UOM ('WSET') instead of inventory UOM ('EA') in 'Reserve Stock to Order' window. Please refer the attached screenshots.

reserve issue

reserve uom

gilmoskowitz:
+1 for cbde014
merge when UAT question about WSET has been addressed and CI passes (waiting on #2586)

foundation-database/public/tables/coitem.sql
```diff
@@ -0,0 +1,3 @@
+SELECT xt.add_column('coitem', 'coitem_qtyreserved_uom_id', 'INTEGER', NULL, 'public');
+
+comment on column public.coitem.coitem_qtyreserved_uom_id is 'UOM of qtyreserved (same as Item Inv UOM).';
```
gilmoskowitz: Does this new column need to be populated? It should have a foreign key constraint.

PR #2586, opened 2016-01-08
issue #27300 - set up fiscal periods and exchange rates

gilmoskowitz:
This actually does a lot more than set up for 2016 success. A recurring problem when we edit these files is remembering to set sequences for serial columns (usually primary keys) to the correct values, so this PR replaces all of the explicit `setval()` calls with a loop over all such columns that calls `setval()` with the max.

Testing note: All reference databases built with these changes (`masterref` is not) will be affected by this. It affects all of the tables in the PostBooks database. If I've done it wrong then we'll start to get unique constraint violations on primary key columns after a few inserts on the affected tables. The **Fix Serial** utility should fix the problems, but then that was the basis for this change so it's kinda redundant.


sindhupriyaav:

gilmoskowitz:
@davidbeauchamp - did you review this already?

PR #2587, opened 2016-01-13
issue #27362 - Itemcosts filter

mccjim:
Adds filter for item costs by class code
Incident 27362


lib/enyo-x/lib/Font-Awesome
```diff
@@ -1 +0,0 @@
-Subproject commit 729766fd3968585c384837b10614c3f6f9b096b4
```
gilmoskowitz: Please undo this change.

lib/module/rjson
```diff
@@ -1 +0,0 @@
-../../node_modules/rjson
```
gilmoskowitz: Please undo the changes to `rjson`.

PR #2589, opened 2016-01-18
Issue #27365:  Images List metasql

andersdd:
Convert Images list to metasql.  Two examples exist in the metasql which calculate or don't calculate image size.  The perofrmance difference in reference Db was 45 seconds -> 1 second.

### DIFF
--- a/foundation-database/public/tables/metasql/images-list.mql
+++ b/foundation-database/public/tables/metasql/images-list.mql
@@ -3,27 +3,13 @@
 -- Notes: Copyright (c) 1999-2016 by OpenMFG LLC, d/b/a xTuple.
 --        See www.xtuple.com/CPAL for the full text of the software license.
 
--- This metaSQL allows turning off the calculation for image size which
--- on large databases takes a very long time.  To use, edit the metaSQL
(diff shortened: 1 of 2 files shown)


foundation-database/public/tables/metasql/images-list.mql
```diff
@@ -0,0 +1,29 @@
...
+
+-- This metaSQL allows turning off the calculation for image size which
+-- on large databases takes a very long time.  To use, edit the metaSQL
+-- and either use the first (default) sql which includes the size calc, or
+-- the second SQL example which does not 
+
+<? if not exists("doNotShowImageSize") ?>
+-- Calculate Image Size for the List
+SELECT image_id, image_name, image_descrip, LENGTH(image_data) AS image_size, 
+                CASE WHEN nspname='public' THEN ''
```
davidbeauchamp: Instead of duplicating the query, couldn't this be simplified to 
```
<? if exists("doNotShowImageSize") ?>
  0 AS image_size,
<? else ?>
  LENGTH(image_data) AS image_size,
<? endif ?>
```
directly in the query?

davidbeauchamp:
+1

PR #2591, opened 2016-01-21
Issue #27205 Additional trial balance checks.  Resolve ordering issues

andersdd:
If we are doing a 4.9.3 it would be nice to backport these fixes to that.

### DIFF
--- a/foundation-database/public/functions/copyaccountingyearperiod.sql
+++ b/foundation-database/public/functions/copyaccountingyearperiod.sql
@@ -1,10 +1,11 @@
 CREATE OR REPLACE FUNCTION copyaccountingyearperiod(pyearperiodid integer)
-  RETURNS integer AS
-$BODY$
--- Copyright (c) 1999-2014 by OpenMFG LLC, d/b/a xTuple. 
+  RETURNS integer AS $$
+-- Copyright (c) 1999-2016 by OpenMFG LLC, d/b/a xTuple. 
 -- See www.xtuple.com/CPAL for the full text of the software license.
 DECLARE
-  _newYear INTEGER;
+  _newYear 	INTEGER;
+  _periodid 	INTEGER;
+  _r 		RECORD;
 BEGIN
 
   -- First create new fiscal year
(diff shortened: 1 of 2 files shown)


foundation-database/public/functions/copyaccountingyearperiod.sql
```diff
@@ -28,17 +29,25 @@ BEGIN
                       EXTRACT(year FROM period_start + INTERVAL '1 year') || '-' || to_char((period_start + INTERVAL '1 year'), 'Mon'),
                       _newYear, period_quarter, period_number
     FROM period
-    WHERE period_yearperiod_id = pYearPeriodid;
+    WHERE (period_yearperiod_id = pYearPeriodid);
+
+--  Grab last period for reference
+    SELECT max(period_id) into _periodid
+    FROM period
+    WHERE (period_yearperiod_id = _newYear);
```
gilmoskowitz: We want `period_id` for the max period end date, which isn't necessarily the `max(period_id)`.

foundation-database/public/functions/copyaccountingyearperiod.sql
```diff
@@ -28,17 +29,25 @@ BEGIN
...
-    FROM accnt LEFT OUTER JOIN trialbal ON ( (trialbal_accnt_id=accnt_id)  )
-    GROUP BY accnt_id
-    HAVING max(trialbal_id) IS NOT NULL
-    ORDER BY accnt_id;
+  FOR _r IN
+    SELECT DISTINCT ON (trialbal_accnt_id) trialbal_id
+    FROM trialbal
+    JOIN period ON (period_id=trialbal_period_id)
+    WHERE period_id = _periodid
+    ORDER BY trialbal_accnt_id, period_end DESC
```
gilmoskowitz: This seems more complicated than necessary. `_periodid` is a single value and we're not using the `period` table for anything here. Also, it's unclear why the order of `trialbal_accnt_id` is important. Why not
```SQL
PERFORM forwardUpdateTrialbalance(trialbal_id)
  FROM trialbal
 WHERE trialbal_period_id = _periodid;
-- or LOOP over
SELECT trialbal_id FROM trialbal WHERE trialbal_period_id = _periodid
```

foundation-database/public/functions/createaccountingperiod.sql
```diff
@@ -96,6 +96,7 @@ BEGIN
     SELECT DISTINCT ON (trialbal_accnt_id) trialbal_id
     FROM trialbal
     JOIN period ON (period_id=trialbal_period_id)
+    WHERE period_id = _periodid
```
gilmoskowitz: Here, too - seems more complicated than necessary because `_periodid` is a single value.

foundation-database/public/functions/copyaccountingyearperiod.sql
andersdd: Actually we only need to pick one period - I don't think it actually matters whether it is the latest - and given we have just inserted the periods for _newYear it most likely *will* be the max period anyway.

foundation-database/public/functions/copyaccountingyearperiod.sql
andersdd: Actually I think adding WHERE _periodid is incorrect - we want to forwardupdate trial balances for all accounts but starting from the latest period.

foundation-database/public/functions/createaccountingperiod.sql
andersdd: On restrospect, we should remove the WHERE _periodid clause

PR #2592, opened 2016-01-21
Provide more user-friendly error message on UoM conversion errors

andersdd:
Provide meaningful Item Number and UoM text rather than internal ids.


davidbeauchamp:
Magic

PR #2594, opened 2016-01-22
issue #26887 - rework _docinfo() to improve perf and eliminate redundancy

gilmoskowitz:
This approach was suggested by @paladinlogic - query the docass and source tables to get the associations, then retrieve detail from various tables individually. The performance selecting directly from the `docinfo` view is terrible. The performance selecting from the `_docinfo(id, type)` function is much better.

Test results on one large database indicate between 1 and 5 orders of magnitude performance, depending on the particular records.

See accompanying xtuple/qt-client#1072

### DIFF
--- a/foundation-database/manifest.js
+++ b/foundation-database/manifest.js
@@ -47,6 +47,7 @@
     "public/indexes/womatlpost.sql",
 
     "public/types/_docinfo.sql",
+    "public/types/_targetdoc.sql",
     "public/types/achline.sql",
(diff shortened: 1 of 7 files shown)


bendiy:
+1

bendiy:
but the test is broken:
`◦ should be retrievable from the view (but slow): <<ERROR 2016-01-22T17:43:27.951Z>> Unrecoverable exception. `

PR #2606, opened 2016-01-29
issue #27437 - fix build_app regression from issue #26887 changes

gilmoskowitz:


rkannadevara:

PR #2618, opened 2016-02-03
Issue #27478: Resolve API.project inconsistency

andersdd:
Everywhere else refers to status C = 'Completed'

### DIFF
--- a/foundation-database/api/views/project.sql
+++ b/foundation-database/api/views/project.sql
@@ -17,7 +17,7 @@ AS
        WHEN (prj_status = 'O') THEN
          'In-Process'
        WHEN (prj_status = 'C') THEN
-         'Closed'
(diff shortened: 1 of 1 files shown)


sindhupriyaav:

andersdd:
Inside xTuple status 'C' is referred to as 'Completed' not 'Closed'.  Also the INSERT and UPDATE of api.project is expecting the value 'Completed'.   Exporting project information from api.project returns status 'Closed' irrelevantly and if you use that status to load back into the view the load fails.

Doing a SELECT * FROM api.project should return the status text 'Completed' to ensure consistency.

sindhupriyaav:
Hi Anderson,

Thanks for the information. Performed UAT. Issue is fixed.

Regards,
Sindhu Priya

PR #2626, opened 2016-02-10
Add the uuid-ossp extension to core. Switch to using it. Add create_v…

bendiy:
…iew() to core.


bendiy:
CI seems to be broken, closing and resubmitting.

bendiy:
See: https://github.com/xtuple/xtuple/pull/2627

PR #2627, opened 2016-02-10
Create view uuid ossp in core

bendiy:


gilmoskowitz:
What's the benefit of this change? It adds a dependency on a contrib module that every customer will need to install.

bendiy:
The tests I've ran show that the `uuid-ossp` extension's `uuid_generate_v4()` function is much faster than the javascript version we've been using.
https://github.com/xtuple/xtuple/blob/4_10_x/lib/orm/source/xt/functions/js_init.sql#L834-L853

You really notice the speed when "mobilizing" a large database. It adds the `obj_uuid` column to several tables and populates it with a UUID for each row. It has to generate a UUID for 100's of thousands of records. Ask @xtpclark about this speed issue.

`js_init()` is in core and has the `XT.generateUUID` function, but the wrapper for it isn't. It's mobile only:
https://github.com/xtuple/xtuple/blob/4_10_x/lib/orm/source/manifest.js#L38
Any tables that have a `uuid` type column need the stored procedure on the database so they can call the function for the column default. The `plv8` JS version in `js_init()` has to be wrapped in a stored procedure to work.

I need UUIDs for the dashboards. Instead of exposing the JS version for core, I wanted to finally add the `uuid-ossp` function to core like I've been hoping to do for 3 years.

It is in [postgres-contrib](http://www.postgresql.org/docs/9.1/static/contrib.html), which all xTuple Postgres servers will have installed because we already require contrib for pgcrypto support. End users should not need to install anything additional.

gilmoskowitz:
@bendiy - The contrib modules are available as files at the OS level but not installed in the databases. End users don't have to worry about it but every DBA does, whether s/he's an IT person or not. We'll have to add this to xTAU and have a compatibility check in the dashboards Updater package.

PR #2631, opened 2016-02-15
Issue #27418:avoid updating transit sites

garyhgohoos:


rkannadevara:
Performed UAT: On clicking 'Query' button for 'INTRAN' site in 'Update Reorder Level by Item' window observed that 'No Item sites records found' system message is being displayed. Please confirm the behavior.

garyhgohoos:
behavior confirmed.  there is no method to remove transit sites from warehouse cluster.  minor abnormality.

PR #2636, opened 2016-02-16
Issue #23725:do not delete cost element when zero

garyhgohoos:


rkannadevara:
Performed UAT: Unable to reproduce the issue in product version. Observed that actual cost of an element (say 'Material') is being displayed as '0' in 'Item' screen on expiring the bill of materials of average cost item. Observed the same behavior in pull database.

garyhgohoos:
the expected result of expiring ALL of the BOM items is 0 actual Material cost.  before fix the Material cost was deleted

PR #2637, opened 2016-02-18
Resolve Documents issues with sources in non-public schema

andersdd:
@gilmoskowitz These changes are working on my test system.  You might want to check my logic though


gilmoskowitz:
@andersdd - these changes should not be necessary because of the `JOIN regexp_split_to_table(buildSearchPath()...)`. What is it about that join that doesn't work properly?

andersdd:
I was missing the login(true) in my sql testing.

PR #2638, opened 2016-02-19
Issue #26903:additional use of format functions

garyhgohoos:


Panduk:
Performed UAT: Issue is fixed

PR #2642, opened 2016-02-23
Issue #25871:consolidate the formatting of addresses

garyhgohoos:


foundation-database/public/functions/formataddr.sql
```diff
@@ -16,34 +15,124 @@ BEGIN
 
   RETURN _return;
 END;
-$$ LANGUAGE 'plpgsql';
+$$ LANGUAGE plpgsql;
 
-CREATE OR REPLACE FUNCTION formatAddr(TEXT, TEXT, TEXT, TEXT, INTEGER) RETURNS TEXT AS $$
+CREATE OR REPLACE FUNCTION formatAddr(pId INTEGER,
+                                      pType TEXT) RETURNS TEXT AS $$
```
gilmoskowitz: This `formatAddr(pId, pType)` function cannot anticipate future growth of core and extensions. It would be a fine idea if the set of `pType`s were fixed, but it isn't. Changing the existing address columns to `not null default ''` would fix the bug as reported. Adding an overload of `formatAddr()` that took _all_ address fields as separate parameters would allow for anticipated future changes to support country-specific address formats.

Panduk:
Performed UAT: 
1. Observed that for Sales orders, Purchase order, credit memo reports  'City' and 'Postal code' are being displayed in printed form 
2. In quotes,invoice reports  'City' and 'Postal code' are not being displayed in printed form 
3. We are not able to check for transfer order because of the INCIDENT# 26322

garyhgohoos:
close pull request and redesign per comments

PR #2649, opened 2016-03-09
Fixes #27155 Add coitemImportedPOPRbeforetrigger.

bendiy:


rkshapiro:
We tested this change and it fixed #27155.  I noticed that the coitem.sql that was modified includes my changes for managing characteristics on sales order lines, which is great.  You may want to remove the lines I commented out and left my initials on if you are going to commit this change.  Look for --rks.

bendiy:
comments removed. ready to merge.

PR #2650, opened 2016-03-13
Issue #26756: Void Posted Checks in alternate bank account currency 

andersdd:
Void Posted Checks in alternate bank account currency and dates.

For some reason this section of code was commented out with a statement "it does not work in all situations"  however commenting out the code was the cause of the issue #26756.

Unfortunately there is no history of this commenting out in Github so I cannot find the reasons behind it.  Is there a way to check prior history?

### DIFF
--- a/foundation-database/public/functions/voidpostedcheck.sql
+++ b/foundation-database/public/functions/voidpostedcheck.sql
@@ -126,8 +126,8 @@ BEGIN
                                           checkitem_amount * -1.0
                                      ELSE checkitem_amount END,
                                   _p.checkhead_checkdate) AS amount_check,
-                     apopen_id, apopen_doctype, apopen_docnumber, apopen_curr_rate, apopen_docdate,
-                     aropen_id, aropen_doctype, aropen_docnumber,
+                     apopen_id, apopen_doctype, apopen_docnumber, apopen_curr_id, apopen_curr_rate,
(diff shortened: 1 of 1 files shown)


Panduk:
Performed UAT: Issue is fixed

gilmoskowitz:
I just checked the [database-obsolete](https://github.com/xtuple/database-obsolete) repo. Looks like this was commented out to address issue 19634.

andersdd:
@gilmoskowitz Just submitted a patch which appears to handle scenarios in both issues now.

gilmoskowitz:
The code doesn't feel right. I'm merging the latest fix but someone should revisit this section of the code and write a test for it.

- `a - b = (b - a) * -1` so half of the comparisons here seem unnecessary but go back to 2011 https://github.com/xtuple/database-obsolete/commit/7e7b9413ccafb0974c8184e290335af1134171d1
- why is `checkitem_curr_rate` not always correct? https://github.com/xtuple/database-obsolete/commit/eed5f99464ae7aecf58a9bf670ca1216e1378423

andersdd:
@gilmoskowitz the scenario in issue 19634 is a really obscure one relating to three different currencies - the Voucher currency and the payment currency where both are *not* the base currency.  In testing it appears we should not be doing an exchange rate conversion for this scenario which is why the original code was commented out and exchGain set to 0.

PR #2655, opened 2016-03-17
Issue #27681:correctly update bankaccnt_nextchknum when dups skipped

garyhgohoos:


rkannadevara:
Performed UAT: Issue is not observed on QA databases.

gilmoskowitz:
To test this, the expected next check number needs to be forced out of sync with the already-used check numbers.

PR #2658, opened 2016-03-21
XT.Data.fetch refactor.

bendiy:
This is a major refactor to `XT.Data.fetch()` and it's associated `XT.Data.buildClause()`. The main goal of this refactor is to add support to modify ORM list request queries before they are executed so we can add additional parts to the query such as:

1. Modify the columns returned to support things like inventory availability calculations by date. See private-extensions associated PR's [`XM.InventoryAvailability.fetch`](https://github.com/bendiy/private-extensions/blob/828da7ab540322dc12173b5cc23170016c9a7caa/source/inventory/database/source/xm/javascript/inventory_availability.sql#L100-L135) change.
2. Add additional `WHERE` clauses to the query. This is used to enforce visibility on Registration Management's [`XM.ScheduledEvent`](https://github.com/bendiy/registration-management/blob/master/resources/database/source/regmgmt/javascript/regmgmt.schevnt.sql#L61-L72) ORM. Also used in the modification in this PR to [`XM.ItemSitePrivate.fetch`](https://github.com/bendiy/xtuple/blob/4976648d48f4fbcc19fb4a59d55067bb72b3c80e/database/source/xm/javascript/item_site.sql#L89-L173) and private-extensions [`XM.InventoryAvailability.fetch`](https://github.com/bendiy/private-extensions/blob/828da7ab540322dc12173b5cc23170016c9a7caa/source/inventory/database/source/xm/javascript/inventory_availability.sql#L140-L166) and [`XM.TransferOrder.items`](https://github.com/bendiy/private-extensions/blob/828da7ab540322dc12173b5cc23170016c9a7caa/source/inventory/database/source/xm/javascript/transfer_order.sql#L60-L90).
3. Also included is the ability to add `JOIN` clauses.

These features are added by the new methods on `XT.Data`:
 * [`XT.Data.buildColumnsExtensions()`](https://github.com/bendiy/xtuple/blob/4976648d48f4fbcc19fb4a59d55067bb72b3c80e/lib/orm/source/xt/javascript/data.sql#L1561)
 * [`XT.Data.buildJoinExtensions()`](https://github.com/bendiy/xtuple/blob/4976648d48f4fbcc19fb4a59d55067bb72b3c80e/lib/orm/source/xt/javascript/data.sql#L1715)
 * [`XT.Data.buildWhereClauseExtensions()`](https://github.com/bendiy/xtuple/blob/4976648d48f4fbcc19fb4a59d55067bb72b3c80e/lib/orm/source/xt/javascript/data.sql#L1868)

There are two ways to modify the ORM query. 

The first way is used on [`XM.ScheduledEvent`](https://github.com/bendiy/registration-management/blob/master/resources/database/source/regmgmt/javascript/regmgmt.schevnt.sql#L61-L72) is a `StaticExtension`. The extension callback function is attached directly to the parent object, `XT.Data`. Any time any list query to the `XM.ScheduledEvent` ORM is requested, the `StaticExtension` callback is always ran.

The second way is a `DynamicExtension` used in all the other examples above. The extension callback function is only attached to the child object directly: `var data = Object.create(XT.Data);` `data.addWhereClauseDynamicExtension(...)`. It will NOT modify any request to that ORM. The `DynamicExtension` callback is ONLY called during that single execution of the `data.fetch()` query. This is useful for our dispatch functions like the examples above. It prevents those query modification from always happening on that ORM and ensures they only happen when you are in direct control of the execution inside a dispatch function that's using `var data = Object.create(XT.Data);`.

An additional change came out of this refactor to `XT.Data.fetch()`. The old function [`XT.Data.buildClause()`](https://github.com/xtuple/xtuple/blob/a2b73971bad0198a1b872962696fa949318a53c8/lib/orm/source/xt/javascript/data.sql#L36-L618) has been removed because it was a 600 line complete mess. All of the logic in `XT.Data.buildClause()` has been broken up into small helper functions that build each part of the query and support the `StaticExtension` and `DynamicExtension` features.

Another change that came out of this refactor to `XT.Data.fetch()` is to query the `xm` views directly. Previously, `XT.Data.buildClause()` built up a `sql1` query which inspected the ORM and queried the underlying base tables directly and returned a list of `id`s that match the parameters. Then a `sql2` query was ran on the `xm` views directly for just those `id`s. That means each time you made an ORM list query, 2 queries had to run. The refactor to `XT.Data.fetch()` makes use of PostgreSQL's support for [Accessing Composite Types](http://www.postgresql.org/docs/9.3/static/rowtypes.html#AEN7450). Instead of quering the base table and then the view, `XT.Data.fetch()` now just queires the view directly and builds up `WHERE` clauses like ``AND ("billingContact").address.city = $1`` that support all the ways `XT.Data.buildClause()` worked. For example, this query:
```
select xt.get($${
  "username": "admin",
  "nameSpace":"XM",
  "type":"ContactListItem",
  "query":{
    "parameters":[
      {
        "attribute": [
          "account.number",
          "account.name",
          "name",
          "phone",
          "address.city"
         ],
        "operator": "MATCHES",
        "value": "ttoys"
      },
      {
        "attribute": "firstName",
        "operator": "BEGINS_WITH",
        "value": "M"
      }
    ]
  }
}$$);
```
Would generate an `sql1` of:
```
select 
  t1.cntct_id as id 
from xt.cntctinfo t1 
left join public.crmacct jt0 on t1.cntct_crmacct_id = jt0.crmacct_id 
left join public.crmacct jt1 on t1.cntct_crmacct_id = jt1.crmacct_id 
left join public.addr jt2 on t1.cntct_addr_id = jt2.addr_id 
where (
  (jt0.crmacct_number ~* $1 
    or jt1.crmacct_name ~* $1 
    or t1.cntct_name ~* $1 
    or t1.cntct_phone ~* $1 
    or jt2.addr_city ~* $1
  ) 
  and (t1.cntct_first_name ~^ $2)) 
group by t1.cntct_id;
```
With values of:
```
["ttoys", "M"]
```
And an `sql2` query of:
```
select * 
from xm.contact_list_item 
where id in ($1);
```
With values of:
```
[1]
```
The refactor to `XT.Data.fetch()` now generates just one query of:
```
SELECT
  *
FROM xm.contact_list_item
WHERE true
  AND (false
    OR (account).number ~* $1 
    OR (account).name ~* $1 
    OR name ~* $1 
    OR phone ~* $1 
    OR (address).city ~* $1 
  )
  AND "firstName" ~^ $2 
;
```
With values of:
```
["ttoys", "M"]
```
This change should speed list queries up and simplifies the query build process.

Also, `XT.Data.fetch()` used to do a for loop over the returned result set at least 4 times. That has been reduced to just one loop.


database/source/xm/javascript/item_site.sql
```diff
@@ -35,314 +35,144 @@ select xt.install_js('XM','ItemSite','xtuple', $$
...
-          effectiveDate = param.value;
-          break;
-        case "vendor":
-          vendorId = data.getId(data.fetchOrm('XM', 'VendorRelation'), param.value);
-          break;
-        default:
-          result = true;
+        switch (param.attribute) {
+          case "customer":
+            customerId = XT.Data.getId(data.fetchOrm('XM', 'CustomerProspectRelation'), param.value);
```
gilmoskowitz: If `data` means the same thing in the old (line 43) and new (line 39) versions, why change the `getId()` calls from `data` to `XT.Data`?

database/source/xm/javascript/item_site.sql
```diff
@@ -35,314 +35,144 @@ select xt.install_js('XM','ItemSite','xtuple', $$
...
+          case "effectiveDate":
+            effectiveDate = param.value;
+            return false;
+          case "vendor":
+            vendorId = XT.Data.getId(data.fetchOrm('XM', 'VendorRelation'), param.value);
+            return false;
+          default:
+            return true;
         }
-        return result;
```
gilmoskowitz: Functions that return values should always end with a `return` - defensive coding in case the flow control is mishandled.

database/source/xm/javascript/item_site.sql
```diff
@@ -35,314 +35,144 @@ select xt.install_js('XM','ItemSite','xtuple', $$
...
+                        '          FROM itemalias\n' +
+                        '          WHERE true\n' +
+                        '            AND itemalias_number ~^ $' + data.whereLiteralValues.length + '\n' +
+                        '            AND (false\n' +
+                        '              OR itemalias_crmacct_id IS NULL\n' +
+                        '              OR itemalias_crmacct_id = $' + data.whereLiteralValues.push(accountId) + '\n' +
+                        '            )\n' +
+                        '        )\n' +
+                        '    )\n' +
+                        '  )\n';
```
gilmoskowitz: The `\n`s don't look necessary. They clutter the code and make the db log a bit harder to search.

database/source/xm/javascript/item_site.sql
```diff
@@ -35,314 +35,144 @@ select xt.install_js('XM','ItemSite','xtuple', $$
...
+           *   AND (false
+           *     OR number ~^ $1
+           *     OR barcode ~^ $1
+           *     -- Plus any additional param.attributes here.
+           *   )
+           *
+           * Trim the trailing `)\n` and spaces off so we can add the `OR id IN (` alias clause below.
+           */
+          var trimmedClause = whereClause.whereClause.trim();
+          var extnSql = '  ' + trimmedClause.substr(0, (trimmedClause.length - 1)) + /* Remove the `)` from the end. */
```
gilmoskowitz: How do we know
- it's safe to assume `trimmedClause` ends with a `)`?
- it's safe to remove it and tack on an `OR`?

database/source/xm/javascript/item_site.sql
```diff
@@ -35,314 +35,144 @@ select xt.install_js('XM','ItemSite','xtuple', $$
...
-      }
-
-      extra += ") ";
-
-      if (!clause.joins) {
-        clause.joins = '';
+      /* If customer passed, restrict results to item sites allowed to be sold to that customer */
+      if (customerId || shiptoId) {
+        data.addWhereClauseDynamicExtension(nameSpace, type, function (originPayload) {
+          var extSql =  '  -- Added by addWhereClauseDynamicExtension\n' +
```
gilmoskowitz: If you use `/* */` style comments, the `\n`s aren't necessary.

database/source/xm/javascript/item_site.sql
```diff
@@ -35,314 +35,144 @@ select xt.install_js('XM','ItemSite','xtuple', $$
...
+                        '    FROM itemsite\n' +
+                        '    WHERE true\n' +
+                        '      AND itemsite_item_id IN (\n' +
+                        '        SELECT custitem(\n' +
+                        '          $' + data.whereLiteralValues.push(customerId) + ',\n' +
+                        '          $' + data.whereLiteralValues.push(shiptoId) + ',\n' +
+                        '          $' + data.whereLiteralValues.push(effectiveDate) + '\n' +
+                        '        )\n' +
+                        '      )\n' +
+                        '  )\n';
```
gilmoskowitz: `FROM itemsite JOIN custitem()` is preferable to nested `SELECT`s unless there's a demonstrable performance gain.

database/source/xm/javascript/item_site.sql
```diff
@@ -35,314 +35,144 @@ select xt.install_js('XM','ItemSite','xtuple', $$
...
+                        '    WHERE true\n' +
+                        '      AND itemsite_item_id IN (\n' +
+                        '        SELECT\n' +
+                        '          itemsrc_item_id\n' +
+                        '        FROM itemsrc\n' +
+                        '        WHERE true\n' +
+                        '          AND itemsrc_active\n' +
+                        '          AND itemsrc_vend_id = $' + data.whereLiteralValues.push(vendorId) + '\n' +
+                        '      )\n' +
+                        '  )\n';
```
gilmoskowitz: Same here - `JOIN` instead of nest unless there's a good reason.

database/source/xm/javascript/item_site.sql
```diff
@@ -35,314 +35,144 @@ select xt.install_js('XM','ItemSite','xtuple', $$
...
+
+          return extSql;
+        });
       }
 
-      /* public.item is not already joined. Add it here. */
-      if (itemJoinTable === 'sidejoin') {
-        clause.joins = clause.joins + XT.format(' left join item %1$I on t1.%2$I = %1$I.item_id ', [itemJoinTable, backingTypeJoinColumn]);
+      /* If vendor passed and vendor can only supply against defined item sources, then restrict results. */
```
gilmoskowitz: Where's the code to enforce the "vendor can only supply against defined item sources" rule?

lib/orm/source/xt/javascript/orm.sql
```diff
@@ -445,6 +445,60 @@ select xt.install_js('XT','Orm','xtuple', $$
...
 
   /**
+   * Get the ORM property of a path attribute. e.g. `contact.address.city`
+   *
+   * @param orm {Object} The ORM to start from.
+   * @param path {String} The path attribute.
+   * @return {Object} The property of a path attribute.
+   */
+  XT.Orm.getPathProperty = function getPathProperty(orm, path) {
+    if (path.indexOf('.') > -1) {
```
gilmoskowitz: Simplify! The call to `indexOf()` is redundant.

lib/orm/source/xt/javascript/orm.sql
```diff
@@ -445,6 +445,60 @@ select xt.install_js('XT','Orm','xtuple', $$
...
+
+  /**
+   * Get the ORM of a path attribute. e.g. `contact.address.city` returns `AddressInfo` ORM
+   *
+   * @param orm {Object} The ORM to start from.
+   * @param path {String} The path attribute.
+   * @return {Object} The ORM of a path attribute.
+   */
+  XT.Orm.getPathOrm = function getPathOrm (orm, path) {
+    var pathParts = (path.indexOf('.') > -1) ? path.split('.') : [path];
```
bendiy: Realized `indexOf` is redundant here, but not above.
`var pathParts = path.split('.');`

lib/orm/source/xt/javascript/orm.sql
bendiy: Did you mean below?
This check is to allow this function to get the ORM if `path` is a true path or just a single column. `contact`. If it's just `contact`, we fall back on `XT.Orm.getProperty()`.

database/source/xm/javascript/item_site.sql
bendiy: It can be assumed because `keySearchAttrMap` is an array of columns, which will run through [`_buildArray()`](https://github.com/bendiy/xtuple/blob/4976648d48f4fbcc19fb4a59d55067bb72b3c80e/lib/orm/source/xt/javascript/data.sql#L1959-L1998) and generate that. This `itemalias` support is tricky. I could add a check that it begins with `AND (` and ends in `)` before tacking on the `OR in IN(`. Any other ideas?

lib/orm/source/xt/javascript/orm.sql
gilmoskowitz: I mean here. Don't ask if you have to `split()`, just do it. Then check the result:
```javascript
var pathParts = path.split('.');
if (pathParts.length == 0) {
  return XT.Orm.getProperty(orm, path);
}
var parentProp = ...
```
This eliminates a function call, puts the simple case at the top, removes a level of indentation for the bulk of the function, and should have about the same run time.

database/source/xm/javascript/item_site.sql
bendiy: Good catch.

I switched this to `XT.Data` early on in the refactor, but didn't switch it back.

Now that I look at it, I think `XT.Data` is correct and I'm going to switch `data.fetchOrm('XM',...` to `XT.Data.fetchOrm('XM','. In this case, `getId()` is a completely separate query operation and has nothing to do with `data.fetch` query. Like a static function on a class, `getId()` should be ran without the `this` context of `data.fetch`. With the current code, it will work fine either way, but the intention here is `getId()` would be a static function on the `XT.Data` class, but `fetch()` is a method on the `data` object that needs it's `this` context to work.

database/source/xm/javascript/item_site.sql
davidbeauchamp: If they help format the query in the debug log I vote for leaving them in, when a MetaSQL query fails it puts a human readable query in the debug log that makes it easy to parse (instead of one long line you have to manually manipulate into place)

gilmoskowitz: I'll defer to folks who work in the field on this.

bendiy: I prefer having new lines in the logs and debug output so I can immediately read and understand they query without having to copy it to a text editor and reformat it every time. Without the `\n`, this will end up being:
```
    OR id IN (      SELECT        itemsite_id AS id      FROM itemsite      WHERE true        AND itemsite_item_id IN (          SELECT            itemalias_item_id          FROM itemalias          WHERE true            AND itemalias_number ~^ $1            AND (false              OR itemalias_crmacct_id IS NULL              OR itemalias_crmacct_id = $2            )        )    )  );
```
in logs and debug output which I don't find very helpful.

MetaSQL preserves the new lines in logs and debug output, so this is consistent with that.

gilmoskowitz: Understood. I prefer having it all on one line so I can `grep` or `sort` the log and find entire queries rather than fragments. Either way is fine.

database/source/xm/javascript/item_site.sql
bendiy: Right. This is what the old code did, but it is missing a check for `vend_restrictpurch`. I'll add it in the next commit.

gilmoskowitz:
Comments about https://github.com/bendiy/xtuple/blob/xtdata-cleanup/lib/orm/source/xt/javascript/data.sql :
- The syntax `myNewFunction: function myNewFunction (...) {...}` is legal but redundant. What happens if the two get out of sync?
- The benefit of splitting `buildQuery()` and `getQueryParts()` is unclear
- The benefit of `buildSelect()` is unclear
- The asymmetry in handling static vs. dynamic columns in `buildColumnsExtensions()` seems unnecessary
- `buildJoinExtensions()` can probably be simplified by concatenating the two arrays
- The switch in `mapOperator()` can be simplified by combining a handful of cases into `operator = ' ' + operator + ' '; break;`
- `_buildPath()` and `_buildNormal()` could simply return arrays of clause strings. The caller knows whether to `join(' AND ')` or `join(' OR ')` them. That would reduce the occurrence of `true AND` and `false OR` and `replace()` calls.
- In `buildOrderBy()`, if `orderBy` is initialized to an empty array when the payload has no `query`, the check `if (orderBy)` becomes unnecessary.

bendiy:
* The syntax myNewFunction: function myNewFunction (...) {...} is legal but redundant. What happens if the two get out of sync?

This is a coding style I read about recently. I'm creating a `Named function expression`. By naming the function, javascript debuggers and stack traces will show the name I gave it instead of `<anon>` or `(?)()`. You are right, they can get out of sync if you are not careful. It's a recommended style to always name your functions.

You can read more about it here:
https://kangax.github.io/nfe/

bendiy:
- The benefit of splitting `buildQuery()` and `getQueryParts()` is unclear
- The benefit of `buildSelect()` is unclear

By splitting `buildQuery()` and `getQueryParts()`, it is possible for code somewhere else to call `getQueryParts()` directly and then change the parts of a new the query easily. 

The reason for splitting these is if we need to re-implemented `XT.Data.fetch()` somewhere in the future. I was able to refactored all of the places that re-implemented `XT.Data.fetch()` in this PR to use it directly. When I started, I wasn't sure if I would be able to do that. 

With two separate functions, you can just get the query parts and then modify them before building a string. It's not needed now, but I think the added flexibility could be useful.

```
myCustomFetch = function (payload) {
  ...
  var queryParts = data.getQueryParts(payload);
  queryParts.from = "some change here";
  
  var sql = queryParts... // join strings together and execute query.
  ...
}
```

The same goes for `buildSelect()`. It's not needed, but we might want to do a `SELECT DISTINCT` somewhere and this allows that.

jcarlin:
+1

bendiy:
@gilmoskowitz 
- The asymmetry in handling static vs. dynamic columns in `buildColumnsExtensions()` seems unnecessary
- `buildJoinExtensions()` can probably be simplified by concatenating the two arrays

I've switch these functions around to add a comment to the query in the buildXXX function instead of relying on the developer to include the comment in the custom `data.fetch` dispatch functions. This ensure the comment is always there and should reduce confusion on where/how the query clause got there. Since the comment is different between the static and dynamic versions, it makes since to have them split out separately. I could make the comment more generic, `-- Added by buildColumnsExtensions`, which would allow me to combine them. Which would you prefer?
```
      if (this.joinClauseStaticExtensions[payload.nameSpace] &&
          this.joinClauseStaticExtensions[payload.nameSpace][payload.type]) {

        var staticExtensions = this.joinClauseStaticExtensions[payload.nameSpace][payload.type];
        staticExtensions.forEach(function (extFunc) {
          joinsClauseExtensions += '-- Added by addJoinClauseStaticExtension\n';
          joinsClauseExtensions += extFunc.call(self, payload);
        });
      }

      if (this.joinClauseDynamicExtensions &&
          this.joinClauseDynamicExtensions[payload.nameSpace] &&
          this.joinClauseDynamicExtensions[payload.nameSpace][payload.type]) {

        var dynamicExtensions = this.joinClauseDynamicExtensions[payload.nameSpace][payload.type];
        dynamicExtensions.forEach(function (extFunc) {
          joinsClauseExtensions += '-- Added by addJoinClauseDynamicExtension\n';
          joinsClauseExtensions += extFunc.call(self, payload);
        });
      }
```
OR
```
      var joinExtensions = [];
      if (this.joinClauseStaticExtensions[payload.nameSpace] &&
          this.joinClauseStaticExtensions[payload.nameSpace][payload.type]) {

        joinExtensions.push(this.joinClauseStaticExtensions[payload.nameSpace][payload.type]);
      }

      if (this.joinClauseDynamicExtensions &&
          this.joinClauseDynamicExtensions[payload.nameSpace] &&
          this.joinClauseDynamicExtensions[payload.nameSpace][payload.type]) {

        joinExtensions.push(this.joinClauseDynamicExtensions[payload.nameSpace][payload.type]);
      }
      
      joinExtensions.forEach(function (extFunc) {
        joinsClauseExtensions += '-- Added by buildJoinExtensions\n';
        joinsClauseExtensions += extFunc.call(self, payload);
      });
```

gilmoskowitz:
I prefer the second option - shorter and easier to read.

PR #2663, opened 2016-03-28
XT.Data cleanup feedback from PR

bendiy:


lib/orm/source/xt/javascript/orm.sql
```diff
@@ -452,22 +452,23 @@ select xt.install_js('XT','Orm','xtuple', $$
...
-      var parentProp = XT.Orm.getProperty(orm, pathParts[0]);
+    var pathParts = path.split('.');
+    if (pathParts.length === 1) {
+      return XT.Orm.getProperty(orm, path);
+    }
 
-      if (parentProp.toOne || parentProp.toMany){
-        var childType = parentProp.toOne ? parentProp.toOne.type : parentProp.toMany.type;
-        var childOrm = XT.Data.fetchOrm(orm.nameSpace, childType);
+    var pathParts = path.split('.');
```
gilmoskowitz: We just did this so there's no need to `split()` again.

bendiy: fixed

PR #2665, opened 2016-03-30
issue #26767 - list prices

garyhgohoos:


foundation-database/public/functions/calcipsitemprice.sql
```diff
@@ -0,0 +1,35 @@
...
+              WHEN ((ipsitem_type = 'M') AND fetchMetricBool('Long30Markups')) THEN
+               (_item.invcost / (1.0 - ipsitem_discntprcnt) + ipsitem_fixedamtdiscount)
+              WHEN (ipsitem_type = 'M' AND fetchMetricBool('WholesalePriceCosting')) THEN
+               (_item.item_listcost + (_item.item_listcost * ipsitem_discntprcnt) + ipsitem_fixedamtdiscount)
+              WHEN (ipsitem_type = 'M') THEN
+               (_item.invcost + (_item.invcost * ipsitem_discntprcnt) + ipsitem_fixedamtdiscount)
+              ELSE 0.00
+         END INTO _result
+  FROM ipsiteminfo
+  WHERE (ipsitem_id=pIpsitemid);
```
gilmoskowitz: `_result` is ignored after we retrieve it.

foundation-database/public/functions/calcipsitemprice.sql
```diff
@@ -0,0 +1,35 @@
...
+
+  IF ( SELECT ((metric_value='t') AND packageIsEnabled('xtmfg'))
+         FROM metric
+        WHERE(metric_name='Routings') ) THEN
```
gilmoskowitz: The `IF` can be simplified by calling `fetchMetricBool()` directly.

foundation-database/public/functions/calcipsitemprice.sql
garyhgohoos: not sure if i need this function yet.  still need to work on pricing reports and may use there.  will delete if don't use.

foundation-database/public/tables/metasql/pricelist-detail.mql
```diff
@@ -152,7 +156,7 @@ FROM (
            <? value('listPrice') ?> AS type,
            -1 AS invqty,
            -1 AS qtybreak,
-           (item_listprice - (item_listprice * cust_discntprcnt)) AS price,
+           (<? value('item_listprice') ?> - (<? value('item_listprice') ?> * cust_discntprcnt)) AS price,
```
gilmoskowitz: This query now requires that a price be passed in? That doesn't sound right.

foundation-database/public/tables/metasql/prices-detail.mql
```diff
@@ -400,7 +400,7 @@ FROM (
              item_id AS itemid, 0 AS sourcetype,
              <? value("listPrice") ?> AS schedulename, <? value("na") ?> AS type,
              '' AS typename,
-             -1 AS qtybreak, item_listprice AS price, baseCurrId() AS curr_id
+             -1 AS qtybreak, <? value("list") ?> AS price, baseCurrId() AS curr_id
```
gilmoskowitz: Same here. Why pass in a list price instead of calculating it? Passing it in is faster but limits the contexts in which the MQL can be used.

PR #2668, opened 2016-04-11
issue #27777 - combine updates and disable triggers

gilmoskowitz:
These changes improve performance of this step of the upgrade by ~10^3 in one customer database.

~~Do not merge until @xtpclark has a chance to test in another.~~ Test in a large customer database where this was initially observed shows significant performance improvement: failure after 8 hours is now success after 25 min.

### DIFF
--- a/foundation-database/public/patches/fixaddress.sql
+++ b/foundation-database/public/patches/fixaddress.sql
@@ -2,17 +2,27 @@ do $$
 declare
   ary     TEXT[];
   col     TEXT;
+  setCol  TEXT[];
+  altCol  TEXT[];
 begin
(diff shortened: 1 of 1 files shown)


andersdd:
@gilmoskowitz  The code looks fine to me bar one question.   I have had issues in the past with disabling and then enabling triggers inside a transaction but that was possibly on an older PG version.

gilmoskowitz:
@andersdd - Good point. Unfortunately we don't have a mechanism for pre-update/post-update tasks outside the updater and build_app transactions. In the database I tested with, 10^2x improvement came from disabling the triggers and only 10x (and avoiding errors!) from combining the updates. Discussions on various PostgreSQL mailing lists indicate this approach is supposed to work.

PR #2671, opened 2016-04-13
fix problem building 4.10.0-beta ref databases using new plv8 pkg

gilmoskowitz:
I encountered a problem building the reference databases for the 4.10.0-beta release. It was caused by a problem with a new plv8 package for Debian/Ubuntu published a week or two ago. Using an older plv8 package works fine.


gilmoskowitz:
@davidbeauchamp - could you please review this? looks like a squash-merge might in order here!

scripts/install_xtuple.sh
```diff
@@ -145,8 +155,7 @@ install_packages() {
     sudo add-apt-repository -y "deb http://ftp.debian.org/debian wheezy-backports main"
   fi
   sudo add-apt-repository -y "deb http://apt.postgresql.org/pub/repos/apt/ ${DEBDIST}-pgdg main"
-  sudo wget -qO- https://www.postgresql.org/media/keys/ACCC4CF8.asc | sudo apt-key add -
```
davidbeauchamp: Is this really not necessary? I don't think PGDG's key is in apt's key repo by default unless its put there by TravisCI

gilmoskowitz: Travis-CI wouldn't pass if this line was in there. `wget` failed to retrieve the file, complaining in essence about missing certificates.

PR #2673, opened 2016-04-15
Issue #27763: Void Cash Receipt did not work for Customer Groups

andersdd:


lokeshsh:

PR #2674, opened 2016-04-18
Fix etag getVersion bug. Cast for in ARRAY. Get column extensions working.

bendiy:


lib/orm/source/xt/javascript/data.sql
```diff
@@ -2471,6 +2501,9 @@ select xt.install_js('XT','Data','xtuple', $$
             result.etags[result.data[i][nkey]] = this.getVersion(this.orm, result.data[i][pkey]);
           }
         }
+
+        /* The sanitizeRow() method will remove the `pkey`, `id`. This must be after etags are handled. */
+        this.sanitizeRow(payload.nameSpace, payload.type, result.data[i], payload);
```
bendiy: @gilmoskowitz This fixes the cause of the `WARNING:  Insert duplicate key error on XT.getVersion.` error we've been seeing.

PR #2676, opened 2016-04-19
Issue #27434: Fix incorrect block count in generated ACH file.

davidbeauchamp:


sindhupriyaav:
Performed UAT: Observed that '1' is displayed instead of '6' in the 6th row of ach file as in the below screenshot. Please confirm the behavior.

davidbeauchamp:
That is correct, that number should be the total lines in the file divided by 10, so in your screenshot you have 6 lines of data, 4 lines of padding, and (10/10) == 1.

foundation-database/public/functions/formatachchecks.sql
```diff
@@ -316,12 +319,21 @@ BEGIN
...
   _rowcount := _rowcount + 1;
+  -- Issue #27434 need to provide block count instead of row count
+  -- In order to do so, must figure out how many rows the BLOCKFILL loop
+  -- below will return to get the final row count, then divide that row count
+  -- by the block size, which in NACHA is 10.
+  IF (_rowcount % BLOCKSIZE) = 0 THEN
+    _blockcount := _rowcount/BLOCKSIZE;
+  ELSE
+    _blockcount := ((BLOCKSIZE - (_rowcount % BLOCKSIZE)) + _rowcount)/BLOCKSIZE;
+  END IF;
```
gilmoskowitz: This can be simplified to `_blockcount := ceiling(_rowcount / BLOCKSIZE);`

davidbeauchamp: the second part of conditional? or the whole conditional block?

gilmoskowitz: The whole 5 lines. See the definition of `ceiling()` at http://www.postgresql.org/docs/current/static/functions-math.html

davidbeauchamp: That is not what I am trying to accomplish, though. NACHA says the ACH file has to be in blocks of 10 lines. so at the time of writing the last data line, I have to figure out if I am already divisible by 10 and if not, how many lines will I have to add in order to be divisible by 10. The total number of lines+padding divided by 10 gives the total number of "blocks" (lines of 10) in the file, which is what the field is asking for and what the bug report is about. 

so for instance, if I have 6 total lines of data, as in the example in the bug report. ceiling(6 / 10) gives back 0, which is incorrect. Same if I use larger number, ceiling(36/10) gives back 3...

In my conditional, the first check is to see if its already divisible by 10, and if it is, than we have our block count already and we know the next loop that does padding will not add any rows. In the second part, I am:

Getting the remainder of rowcount % 10: ie 6 % 10 = 6
Subtracting that figure from the blocksize: 10 - 6 = 4
Adding that answer to the original rowcount in order to pad it to the next whole 10s: 6+4 = 10
Then dividing that whole thing by the block size: 10/10 = 1.
For a total of one block, putting 000001 in the field for block count in the ACH file. 

So if we had 33 rows:
(( 10 - (33 % 10)) + 33)/10 = 4
for a total of 4 blocks of 10.

gilmoskowitz: The behavior of `ceil` depends on the type of its input, which in turn depends on the type coercion of `/`:
```
# select ceil(6.0/10.0) as ff, ceil(6/10.0) as if, ceil(6.0/10) as fi, ceil(6/10) as ii;
 ff | if | fi | ii 
----+----+----+----
  1 |  1 |  1 |  0
(1 row)
```

gilmoskowitz: [ And `/:` is appropriate punctuation there ]

PR #2687, opened 2016-04-26
issue #27692 - Properly handle inventory history for work orders

ChristopherCotnoir:
For work orders with multiple instances of the same item, have the inventory history QOH before and after columns show the correct amounts.

### DIFF
--- a/foundation-database/public/functions/issuewomaterial.sql
+++ b/foundation-database/public/functions/issuewomaterial.sql
@@ -73,15 +73,15 @@ BEGIN
 END;
 $$ LANGUAGE 'plpgsql';
 
-CREATE OR REPLACE FUNCTION issueWoMaterial(INTEGER, NUMERIC, INTEGER, TIMESTAMP WITH TIME ZONE) RETURNS INTEGER AS $$
+CREATE OR REPLACE FUNCTION issueWoMaterial(INTEGER, NUMERIC, INTEGER, TIMESTAMP WITH TIME ZONE, NUMERIC) RETURNS INTEGER AS $$
 -- Copyright (c) 1999-2014 by OpenMFG LLC, d/b/a xTuple. 
 -- See www.xtuple.com/CPAL for the full text of the software license.
 BEGIN
-  RETURN issueWoMaterial($1, $2, $3, $4, NULL);
+  RETURN issueWoMaterial($1, $2, $3, $4, NULL, $5);
 END;
 $$ LANGUAGE 'plpgsql';
 
-CREATE OR REPLACE FUNCTION issueWoMaterial(INTEGER, NUMERIC, INTEGER, TIMESTAMP WITH TIME ZONE, INTEGER) RETURNS INTEGER AS $$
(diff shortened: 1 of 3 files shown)


lokeshsh:

foundation-database/public/functions/issuewomaterial.sql
```diff
@@ -73,15 +73,15 @@ BEGIN
 END;
 $$ LANGUAGE 'plpgsql';
 
-CREATE OR REPLACE FUNCTION issueWoMaterial(INTEGER, NUMERIC, INTEGER, TIMESTAMP WITH TIME ZONE) RETURNS INTEGER AS $$
+CREATE OR REPLACE FUNCTION issueWoMaterial(INTEGER, NUMERIC, INTEGER, TIMESTAMP WITH TIME ZONE, NUMERIC) RETURNS INTEGER AS $$
```
gilmoskowitz: This needs to maintain backwards compatibility (e.g. see lines 29 and 72 above). Either add a distinct 5-parameter overload or make this particular overload take a default value of 0.0 for the new parameter.

foundation-database/public/functions/postinvtrans.sql
```diff
@@ -1,4 +1,5 @@
-CREATE OR REPLACE FUNCTION postInvTrans(pItemsiteId    INTEGER,
+CREATE OR REPLACE FUNCTION postInvTrans(pPrevQty       NUMERIC,
```
gilmoskowitz: Please don't add a new parameter at the beginning of a call sequence. This breaks a long-standing convention. Please put `pPrevQty` at the end and make it optional via a default value. You may need to add `DROP FUNCTION IF EXISTS` for the original call sequence to avoid errors with ambiguous function calls.

The rationale for adding at the end, even though earlier in the sequence might be more semantically meaningfuil, is that we don't know all of the callers of a stored procedure. They might be in other repositories or in customizations.

lokeshsh:

foundation-database/public/functions/postinvtrans.sql
```diff
@@ -12,6 +13,7 @@ CREATE OR REPLACE FUNCTION postInvTrans(pItemsiteId    INTEGER,
                                         pTimestamp     TIMESTAMP WITH TIME ZONE
                                                        DEFAULT CURRENT_TIMESTAMP,
                                         pCostOvrld     NUMERIC DEFAULT NULL,
+                                        pPrevQty       NUMERIC DEFAULT NULL,
```
gilmoskowitz: Please put the new `pPrevQty` after the existing `pInvhistid` parameter.

foundation-database/public/functions/postinvtrans.sql
```diff
@@ -123,8 +125,8 @@ BEGIN
       invhist_series )
   SELECT
     _invhistid, itemsite_id, pTransType, _timestamp,
-    pQty, itemsite_qtyonhand,
-    (itemsite_qtyonhand + (_sense * pQty)),
+    pQty, (itemsite_qtyonhand + (_sense * pPrevQty)),
+    (itemsite_qtyonhand + (_sense * pQty) + (_sense * pPrevQty)),
```
gilmoskowitz: Please `coalesce(pPrevQty, 0.0)` to ensure `itemsite_qtyonhand` doesn't go `NULL`

PR #2691, opened 2016-04-27
Issue #27775:use unformatted value to calc total

garyhgohoos:


lokeshsh:
Performed UAT: Observed that in Detail field on the footer of the reporf of QOH, column name is changed to "value".
Please confirm the behavior

gilmoskowitz:
The column name change to `value` is correct.

PR #2701, opened 2016-05-12
Issue #27913:show list price schedule name on priceList

garyhgohoos:


foundation-database/public/functions/listpriceschedule.sql
```diff
@@ -0,0 +1,93 @@
...
+
+--  If item is exclusive then list list price does not apply
+  IF (_item.item_exclusive) THEN
+    RETURN '';
+  END IF;
+
+  RETURN '';
+
+END; $$
+  LANGUAGE plpgsql;
```
gilmoskowitz: This function is nearly identical to `listprice()`. Please move the commonalities to a 3rd function so we don't have to worry about code maintenance (particularly the big query that gets the pricing schedule).

garyhgohoos:
i simplified as much as possible without going to creating a new data type.  the logic to retrieve the appropriate pricing schedule is already in a couple of functions and should not be a problem.

PR #2706, opened 2016-05-24
Issue #27962 Add trigger to set shiptoinfo defaults if they aren't provided.

bendiy:
We added default configs for these to `Sales > Setup > Customer Defaults` a while ago. The PR uses those metrics to set defaults.


rkannadevara:

bendiy:
@rkannadevara This can be tested in the Qt Client by adding or editing a Customer's Ship To. Leave the following fields blank.
ship-to 2016-05-25 09 39 59

The `Customer Defaults` must be configured:
setup 2016-05-25 09 41 26

PR #2710, opened 2016-05-30
Issue #27740: Prevent Db editing of base currency after it has been set.

andersdd:


rkannadevara:
Performed UAT: Unable to edit the base currency from table level. Issue is fixed.

PR #2713, opened 2016-06-01
Issue #27976:honor work days when rescheduling children

garyhgohoos:


rkannadevara:
Performed UAT: Observed that on changing both start date and due date to non-working days, sub-assemblies are changed to working days as expected.

PR #2720, opened 2016-06-08
Issue #27919:fix check for shipping kit components complete

garyhgohoos:


rkannadevara:
Tested in xTuple 4.10.0 Beta2 nightly build: Issue is not observed on QA databases and able to dropship the purchase order and ship the kit item successfully without any erros

foundation-database/public/functions/shipshipment.sql
```diff
@@ -76,22 +76,18 @@ BEGIN
...
+                      JOIN shipitem ON (shipitem_orderitem_id=coitem_id AND shipitem_shiphead_id=pshipheadid)
+                WHERE ((coitem_status NOT IN ('C','X'))
+                  AND  (item_type != 'K')
+                  AND  (coitem_cohead_id=_shiphead.shiphead_order_id)
+                  AND  (coitem_subnumber <> 0))
+                GROUP BY coitem_id, coitem_qtyshipped, coitem_qtyord, coitem_qtyreturned
+      LOOP
+        IF (_k.remain > 0) THEN
+          RAISE EXCEPTION 'Kit component item not shipped complete.  Kits must be shipped and shipped complete or closed on the order.';
+        END IF;
```
gilmoskowitz: This could be rewritten as
```
IF EXISTS(SELECT 1
            FROM coitem join shipitem ...
           WHERE ...
             AND coitem_qtyord > COALESCE(...)) THEN
  RAISE EXCEPTION ...
END IF;
```
The benefit of using an outer join to get the item is unclear as `item_type != 'K'` will return false if there is no matching itemsite record, thereby filtering out these cases anyway.

The exception message should include the `[xtuple: functionname, errorcode]` suffix.

garyhgohoos:
not sure how to rewrite query as "IF EXISTS" and move the summary and grouping to the WHERE clause.  reacted to other comments

PR #2724, opened 2016-06-10
Issue #28055:change query column from _balance to balance

garyhgohoos:


lokeshsh:
Performed UAT : Observed that query column name is changed from '_balance'  to 'balance' in 'issallbalancetoshipping' function.

PR #2726, opened 2016-06-13
Resolve sequence issue with tax demo data

andersdd:
For some reason the starting sequence was 14 and if you created more than 2 tax codes, you hit a duplicate key error.


andersdd:
Tests failing due to hardcoded tax ids - going to try another method

PR #2740, opened 2016-06-24
Issue #28108:populate cohist_promisedate

garyhgohoos:


foundation-database/public/functions/postinvoice.sql
```diff
@@ -141,7 +141,7 @@ BEGIN
 
 --  March through the Non-Misc. Invcitems
   FOR _r IN SELECT *
-            FROM invoiceitem
+            FROM invoiceitem LEFT OUTER JOIN coitem ON (coitem_id=invcitem_coitem_id)
```
gilmoskowitz: This is going to break in a web-enabled database because both `invoiceitem` and `coitem` have columns named `obj_uuid`.

sindhupriyaav:

gilmoskowitz:
Testing the fix requires either looking at the cohist table with psql or pgadmin, or adding a column to an existing Sales History display and associated metasql statement as described in the original bug report.

PR #2742, opened 2016-06-24
Issue #28072:fix comment source

garyhgohoos:


sindhupriyaav:
Performed UAT: Observed the line
 " VALUES (COALESCE(new.date, now()), 'I'::text, getitemid(new.item_number::text), COALESCE(new.username, geteffectivextuser()), getcmnttypeid(new.type), new.text);" 
is displayed in 'api.itemcomment' view of Pull Request database. Please confirm the behaviour.

gilmoskowitz:
@sindhupriyaav -  That view definition looks correct. To test more thoroughly, you should be able to insert a comment using the view then retrieve it again. The bug is an inconsistency between selecting from the view and inserting into the view, where inserts used the wrong type.

PR #2747, opened 2016-06-30
issue #28131 Added missing condition

ChristopherCotnoir:


rkannadevara:
Performed UAT: Tested in distribution empty database created from pull request branch and observed that issue still exists. Created sales order is not being displayed in 'Open Sales Orders' screen.

ChristopherCotnoir:
Run queries

SELECT * FROM cohead;

and

SELECT * FROM cohead JOIN site() ON (cohead_warehous_id=warehous_id)
WHERE NOT EXISTS(SELECT 1
                FROM coitem
                JOIN itemsite on (coitem_itemsite_id=itemsite_id)
                LEFT OUTER JOIN site() on (itemsite_warehous_id=warehous_id)
               WHERE coitem_itemsite_id IS NOT NULL
                AND warehous_id IS NULL
                AND coitem_cohead_id=cohead_id) );

as the database user you log into xtuple with for testing.

If the first query returns results but the second query does not, then this is intended behavior. There was a security hole in the application wherein sales orders that are either associated with a site the user does not have access to or have line items associated with a site the user does not have access to would appear in the Open Sales Orders screen, when they should be hidden. If the second query returns no results but the first one does, this means that all of the sales orders that are created by the scripts you use to populate the empty databases are associated with a site your user does not have access to. If this is the case, they should never have appeared and only ever did due to a security exploit, and you should tweak the scripts to ensure the user you log in as has access to all the sites associated with any sales orders you want to see.

rkannadevara:
Hi Christopher,

After running the 1st query on pull request database returns the results, but running 2nd query returns the following error:

ERROR:  syntax error at or near ")"
LINE 8: AND coitem_cohead_id=cohead_id) );
                                        ^
********** Error **********

ERROR: syntax error at or near ")"
SQL state: 42601
Character: 312

Also observed that, results are displayed after removing the ')' from line 8 and rerunning the 2nd query.

Thanks,
Rama krishna Kiran

PR #2749, opened 2016-07-01
issue #25167 Fixed boolean typos and skipped share_user cache refresh when not needed

ChristopherCotnoir:


foundation-database/public/functions/checkprivilege.sql
```diff
@@ -5,12 +5,23 @@ DECLARE
   pPrivilege ALIAS FOR $1;
   _result TEXT;
 BEGIN
+  RETURN checkPrivilege(pPrivilege, getEffectiveXtUser());
+END;
+$$ LANGUAGE 'plpgsql';
+CREATE OR REPLACE FUNCTION checkPrivilege(text, text) RETURNS BOOLEAN STABLE AS $$
```
bendiy: Just make this an overload function.
```
CREATE OR REPLACE FUNCTION checkPrivilege(text, text DEFAULT getEffectiveXtUser())...
```

Then `DROP FUNCTION IF EXISTS checkPrivilege(text);`

lib/orm/source/xt/javascript/shareusers.sql
```diff
@@ -79,13 +79,21 @@ select xt.install_js('XT','ShareUsers','xtuple', $$
...
+   * the ORM privilege check is not going to do a share access check
+   * on them, so those types in the share_users table do not need
+   * to be cached.
    *
    * @param {string} The username of the user to refresh the cache for.
    */
   XT.ShareUsers.refreshCacheUser = function(refreshObjUser) {
-    var refreshObjSql = 'select xt.refresh_cache_share_users_user($1)';
+    var privCheckSql = 'SELECT checkPrivilege($2, $1) AND checkPrivilege($3, $1) AND checkPrivilege($4, $1)';
+    if(!(plv8.execute(privCheckSql, [refreshObjUser, 'MaintainAllCRMAccounts', 'MaintainCustomMasters', 'MaintainShiptos']))) {
```
gilmoskowitz: The idea is right but the handling of `plv8.execute()` is not - it [returns an array of objects](http://pgxn.org/dist/plv8/doc/plv8.html#plv8.execute.sql.args.), which in this case will contain a single element.

PR #2751, opened 2016-07-07
issue #25989 Update GL when Cost Category changes

ChristopherCotnoir:


ChristopherCotnoir:
Requires xtuple/qt-client#1299.

PR #2756, opened 2016-07-11
Issue #27631: Missing Item attributes when copying an Item

andersdd:


sindhupriyaav:

PR #2758, opened 2016-07-11
issue #28218 - Added Resolution search/column to Incident Workbench

ChristopherCotnoir:


ChristopherCotnoir:
Requires xtuple/qt-client#1306.

PR #2774, opened 2016-07-27
Update Quickstart Db to make more generic and add extra components

andersdd:
Sales Tax,  UoM, Terms


gilmoskowitz:
I'm not sure why CI is failing. When I run all 5 test/models files individually, test/models/location.js and model.js fail with the same error:
```
1)  "before all" hook:
     AssertionError: "Error: \"Error: Error: You must supply a valid Vendor Name.\"" undefined "undefined"
```
and the other 3 pass. If I run them all again individually, they _all_ pass. Stranger still:
```
1$ npm run-script test-build
2$ for i in 1 2 ; do
    for test in test/models/location.js test/models/model.js ; do
        echo $i $test =======================
        node_modules/.bin/mocha -R spec test/lib/login $test
    done
done
```
if you run `2$` enough times, both tests eventually pass.

PR #2778, opened 2016-07-29
Could not view/edit prices on inactive Item Source

andersdd:
Conflict between uses of metaSQL in different areas of the application.


rkannadevara:
Performed UAT: Observed that able to view/edit prices on inactive Item sources and also able to add a new price on inactive item sources.

gilmoskowitz:
@andersdd - does this fix a reported bug?

PR #2780, opened 2016-07-29
issue #26139 Use Inventory Adjustment account instead of P/O Liability Clearing ac…

ChristopherCotnoir:
…count for Material Receipts


rkannadevara:

PR #2785, opened 2016-08-05
issue #28338 - support the search box

gilmoskowitz:


foundation-database/public/tables/metasql/itemSources-detail.mql
```diff
@@ -49,6 +49,15 @@
...
 <? endif ?>
+<? if exists("search_pattern") ?>
+   AND (item_number ~* <? value("search_pattern") ?>
+        OR  (COALESCE(item_descrip1,'') || COALESCE(item_descrip2,'') ~* <? value("search_pattern") ?>)
+        OR  (COALESCE(itemsrc_vend_item_number,'') ~* <? value("search_pattern") ?>)
+        OR  (COALESCE(itemsrc_manuf_name,'') ~* <? value("search_pattern") ?>)
+        OR  (COALESCE(itemsrc_manuf_item_descrip,'') ~* <? value("search_pattern") ?>)
+        OR  (vend_name ~* <? value("search_pattern") ?>)
+        OR  (EXISTS(SELECT itemalias_id FROM itemalias WHERE (itemalias_item_id=item_id) AND (COALESCE(itemalias_number,'') ~* <? value("search_pattern") ?>))) )
+<? endif ?>
```
bendiy: @gilmoskowitz Let's say my `search_pattern` === `''` and any one of these columns is `NULL` on most items. This could return all of those `NULL` items. Is that the expected result?

bendiy:
+1 if that^ use case if fine.

foundation-database/public/tables/metasql/itemSources-detail.mql
bendiy: For example if you have defined some other parameter widget search like the `Item` to narrow the results, but the `search_pattern === ''` opens it back up to more.

gilmoskowitz: This is copied directly from the [distribution version](https://github.com/xtuple/private-extensions/blob/4_10_x/source/distribution/foundation-database/xwd/tables/pkgmetasql/itemSources-detailxwd.mql#L52-L60) of this query.

- If the search_pattern is an empty string then we're saying there's no restriction based on that search_pattern, so returning all of those rows with `NULL` is appropriate.
- If there's another parameter widget search criterion, that filter will be effective because of the `AND (...)` restriction and expression grouping on `search_pattern`.

PR #2787, opened 2016-08-08
issue #28350 - add itemalias check

gilmoskowitz:


rkannadevara:
Performed UAT: Issue still exists. Observed that item records are not displayed in 'Items' list screen on searching with item alias in 'Item Number Pattern'.

garyhgohoos:
test probably failed because testing with enterprise/distribution db that has customized items-detail metasql.  will create pull request under private-extensions

PR #2789, opened 2016-08-10
add states for common countries, fix curr symbols and related, remove cruft

gilmoskowitz:


andersdd:
+1 code review when CI passes.   Checked states, countries, metaSQL and report.  Unicode characters appear intact.  NOTE: These files will require Updater 2.3.0 or greater otherwise the characters are not retained.

PR #2794, opened 2016-08-18
Issue #28408: Correct data type bug in getcompanyid function

andersdd:
Function assumes company is a integer value instead of alphanumeric

### DIFF
--- a/foundation-database/public/functions/getcompanyid.sql
+++ b/foundation-database/public/functions/getcompanyid.sql
@@ -10,8 +10,9 @@ BEGIN
     RETURN -1;
   END IF;
 
-  SELECT accnt_company INTO _returnVal
+  SELECT company_id INTO _returnVal
   FROM accnt
+  JOIN company ON (company_number=accnt_company)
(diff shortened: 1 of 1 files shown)


sindhupriyaav:
Performed UAT: Created company '1A' and executed 'SELECT getcompanyid(1A)' in database. Observed following error. 

ERROR:  syntax error at or near "A"
LINE 2: 1A
         ^
********** Error **********

ERROR: syntax error at or near "A"
SQL state: 42601
Character: 23

davidbeauchamp:
@sindhupriyaav I believe you need to pass in an accnt_id from the accnt table, and it will return the company id, so "SELECT getcompanyid(109)" should return "01" from the demo databases.

andersdd:
@sindhupriyaav @davidbeauchamp  Actually it should return the company internal id so return 1 not '01'

The original error was assuming the accnt_company and company_code where always integers when in fact they are text fields

lokeshsh:
Performed UAT: Issue is fixed

PR #2797, opened 2016-08-19
issue #28348 - add optional active flag field to source table

gilmoskowitz:
@davidbeauchamp - Could you review this please? I only added an active parameter to those source records that have a boolean field for this purpose. Let me know if you want this expanded to cover other cases:
-  `(cohead_status = 'O') as cohead_active`
-  (`lsreg_expiredate > current_date`) as lsreg_active`
- etc.

### DIFF
--- a/foundation-database/public/patches/populate_source.sql
+++ b/foundation-database/public/patches/populate_source.sql
@@ -368,7 +368,7 @@ select createDoctype(21, --pDocAssNum
                      '', --pUi
                      '', --pPriv
                      'Inventory',       -- pModule
-                     'location_usable'  -- pActive
+                     'location_active'  -- pActive
 );
 select createDoctype(22, --pDocAssNum
                      'LS', --pType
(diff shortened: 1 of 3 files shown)


foundation-database/public/patches/populate_source.sql
```diff
@@ -359,7 +367,8 @@ select createDoctype(21, --pDocAssNum
                      '', --pParam
                      '', --pUi
                      '', --pPriv
-                     'Inventory' --pModule
+                     'Inventory',       -- pModule
+                     'location_usable'  -- pActive
```
davidbeauchamp: This one isn't quite right, usable is a state of the inventory within the location, not the location itself. Incident #20423 added the ability to deactivate locations via the column location_active on PR [#1308](https://github.com/xtuple/qt-client/pull/1308)

davidbeauchamp:
+1 with location corrected. Only populating tables with a direct active column satisfies my original need,  people can add inferred activate booleans like your example if desired.

PR #2801, opened 2016-08-23
Issue #27761: Exclude Lot Serial characteristics from unique checks

andersdd:
as these characteristics are required to be retained for traceability purposes


rkannadevara:

PR #2805, opened 2016-08-25
issue #28433 Skip workflow if there is no sales, project, purchase order, etc. type

ChristopherCotnoir:


ChristopherCotnoir:
Related to xtuple/private-extensions#829.

PR #2809, opened 2016-08-30
issue #28447 - processing the manifest must preserve properties

gilmoskowitz:
The manifest files allow specifying either a filename or an object containing the filename and other properties describing how that file should be loaded into the database, such as grade. This auxiliary information was being stripped off. This change allows build_app.js to tweak the file paths while preserving the extra details.


ChristopherCotnoir:
+1

PR #2810, opened 2016-08-31
issue #28462 Make it possible to delete cohead records again

ChristopherCotnoir:


foundation-database/public/trigger_functions/cohead.sql
```diff
@@ -511,6 +511,8 @@ BEGIN
     NEW.cohead_lastupdated := now();
   ELSIF (TG_OP = 'UPDATE') THEN
     NEW.cohead_lastupdated := now();
+  ELSIF (TG_OP = 'DELETE') THEN
+    RETURN OLD;
```
gilmoskowitz: Instead of adding code here, remove code elsewhere. This trigger _function_ no longer deals with `DELETE` so change the trigger itself not to fire on delete.

PR #2812, opened 2016-09-06
issue #28280 - removed transaction block from voucherItem

ChristopherCotnoir:
and fixed split receipt freight functionality


ChristopherCotnoir:
Associated with xtuple/qt-client#1361.

PR #2813, opened 2016-09-07
issue #20404 Replace SELECT * with only necessary columns

ChristopherCotnoir:


ChristopherCotnoir:
Related to xtuple/qt-client#1364.

PR #2820, opened 2016-09-09
issue #28472 Update adjustInvValue return value

sbuttgereit:
Prior to this change public.adjustInvValue(intger, number, integer)
would only return 0 on success or -1 on failure.  This change causes a
successful processing to return the invhist_id value of the
public.invhist record created by the function.  The function still
returns -1 on failure.

### DIFF
--- a/foundation-database/public/functions/adjustinvvalue.sql
+++ b/foundation-database/public/functions/adjustinvvalue.sql
@@ -1,14 +1,13 @@
-CREATE OR REPLACE FUNCTION adjustInvValue(INTEGER, NUMERIC, INTEGER) RETURNS INTEGER AS $$
+CREATE OR REPLACE FUNCTION adjustInvValue(pItemsiteid   INTEGER, 
+                                          pNewValue     NUMERIC, 
+                                          pAccountid    INTEGER) 
+  RETURNS INTEGER AS $$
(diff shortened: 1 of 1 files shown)


davidbeauchamp:
A quick search didn't turn up anything, but does anything call this function that expects a 0 result?

sbuttgereit:
@davidbeauchamp  Short answer is no.  A fuller assessment of impact can be found at https://www.xtuple.org/xtincident/view/features/28472

davidbeauchamp:
Gotcha. Thank you for the contribution, the change looks reasonable to me but will defer to @gilmoskowitz for the merge.

sbuttgereit:
I noticed that _I_ have the merge pull request button as well.... that can't right. @gilmoskowitz you might want to have a GitHub permissions check done for non-maintainer accounts such as mine.

gilmoskowitz:
@sbuttgereit - thank you for the contribution and the security warning! The group you are in should no longer have merge privilege. I'm going to hold off merging until we get 4.10.0RC out the door.

sbuttgereit:
Many thanks @gilmoskowitz (and RWM who sponsored the work).  I will update this pull request with the clean-up mentioned in our email exchange.

PR #2821, opened 2016-09-10
 	issue #28468 Add Gl Acct ID param to invAdjustment

sbuttgereit:
This change accomplishes three tasks per the recorded issue: 1- Drops
the original 6 parameter version of the function; 2- Create a function
with the same code, but an extra parameter for overriding the inventory
adjustment gl account sourced from the itemsite's cost category; 3-
Coalesces the parameter value and the cost category supplied GL account
when passing parameters to the postInvTrans function call.


andersdd:
+1 code review.  Looks good to me.

PR #2829, opened 2016-09-20
issue #28547 Added missing joins and fixed incorrect function arguments

ChristopherCotnoir:


foundation-database/public/tables/metasql/timePhasedSales-detail.mql
```diff
@@ -41,9 +45,9 @@ FROM (
 <? else ?>
        , cohist_qtyshipped
   <? if exists("capacityUnits") ?>
-       * itemcapinvrat(itemsite_item_id)
+       * itemcapinvrat(item_id)
```
gilmoskowitz: There might be the same ambiguity here about which `item_id` column is intended.

PR #2830, opened 2016-09-22
issue #28582 - Update voidapvoucher

mccjim:
Changes test for open vouchers as reported on issue 28582

### DIFF
--- a/foundation-database/public/functions/voidapopenvoucher.sql
+++ b/foundation-database/public/functions/voidapopenvoucher.sql
@@ -77,9 +77,10 @@ BEGIN
   END IF;
 
 --  Check for APApplications
-  SELECT apapply_id INTO _test
-  FROM apapply
-  WHERE (apapply_target_apopen_id=_n.apopen_id)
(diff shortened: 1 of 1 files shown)


rkannadevara:
Performed UAT: Unable to verify the fix. Please provide more information to verify the fix.

foundation-database/public/functions/voidapopenvoucher.sql
```diff
@@ -77,9 +77,10 @@ BEGIN
   END IF;
 
 --  Check for APApplications
-  SELECT apapply_id INTO _test
-  FROM apapply
-  WHERE (apapply_target_apopen_id=_n.apopen_id)
+  SELECT apopen_id INTO _test
+  FROM apopen
+  WHERE (apopen_id=_n.apopen_id 
+  AND apopen_paid != 0)
```
gilmoskowitz: We already retrieved this info when we set `_n`. Instead of `select`ing again, just change the conditional to `IF _n.apopen_paid != 0) THEN`

PR #2823, opened 2016-09-12
issue #16967 Change how recurring items are generated DO NOT MERGE

ChristopherCotnoir:

PR #2831, opened 2016-09-23
issue #16967 - change how recurring items are generated

gilmoskowitz:
replaces xtuple/xtuple#2823
requires xtuple/qt-client#1387


foundation-database/public/functions/createrecurringitems.sql
```diff
@@ -105,10 +112,81 @@ BEGIN
...
+               || '    AND ([schedcol]<date_trunc(''minute'', now())));';
+
+    _maxpaststmt := REPLACE(_maxpaststmt, '[schedcol]', _rt.recurtype_schedcol);
+    _maxpaststmt := REPLACE(_maxpaststmt, '[fulltable]',_rt.recurtype_table);
+    _maxpaststmt := REPLACE(_maxpaststmt, '[table]',
+                          REGEXP_REPLACE(_rt.recurtype_table, E'.*\\.', ''));
+    _maxpaststmt := REPLACE(_maxpaststmt, '[done]',  _rt.recurtype_donecheck);
+    _maxpaststmt := REPLACE(_maxpaststmt, '[limit]', COALESCE(_rt.recurtype_limit,
+                                                     'TRUE'));
+    _maxpaststmt := REPLACE(_maxpaststmt, '$1', _r.recur_parent_id::TEXT);
```
gilmoskowitz: This set of string `replace`ments is better handled by calling the `format()` function. The source string for the replacements might be easier to read if dollar-quoted instead of using single-quotes and concatenation operators:
```
_stmt := format($f$SELECT MAX(%I)::TEXT FROM...
                    WHERE ...
                      AND NOT %s
                      AND ... ;
                $f$,
                _rt.recurtype_schedcol, ..., _rt.recurtype_donecheck, ...);
```

foundation-database/public/functions/createrecurringitems.sql
```diff
@@ -105,10 +112,81 @@ BEGIN
...
+    _updatelast := 'AND [schedcol]=''[last]''';
+
+    _updatelast := REPLACE(_updatelast, '[schedcol]', _rt.recurtype_schedcol);
+
+    IF(_r.recur_style='KeepOne') THEN
+      EXECUTE _maxpaststmt INTO _lastpast;
+      _updatelast := REPLACE(_updatelast, '[last]', _lastpast);
+    ELSE
+      _updatelast := '';
+    END IF;
```
gilmoskowitz: `_updatelast` could be initialized to the empty string and only modified if `KeepOne`. That would reduce the amount of manipulate of that string and concentrate what's left in one block.

foundation-database/public/tables/recur.sql
```diff
@@ -0,0 +1,27 @@
+select xt.create_table('recur', 'public');
```
gilmoskowitz: Could you please remove the now-redundant lines in 440_schema.sql?

gilmoskowitz:
[CI failure](https://travis-ci.org/xtuple/xtuple/builds/162850051#L935):
```
psql:/home/travis/build/xtuple/xtuple/foundation-database/440_schema.sql:109506: ERROR:  relation "recur" does not exist
```

foundation-database/public/functions/createrecurringitems.sql
```diff
@@ -117,7 +186,7 @@ BEGIN
       RAISE DEBUG 'createrecurringitems looping, existcnt = %, max = %, is % between % and %?',
                   _existcnt, _r.recur_max, _next, _r.recur_start, _r.recur_end;
 
-      IF (_next BETWEEN _r.recur_start AND _maxdate) THEN
+      IF (_next BETWEEN _r.recur_start AND _maxdate AND _next >= date_trunc('minute', now())) THEN
```
gilmoskowitz: This change is the crux of the fix, right? The problem is that it doesn't discriminate between high and low frequency jobs. I don't see how the new code distinguishes between jobs that need to run a set number of times and jobs that can catch up on a backlog by themselves.

foundation-database/public/functions/createrecurringitems.sql
```diff
@@ -105,10 +112,72 @@ BEGIN
...
+                    WHERE ((%L=%I_recurring_%I_id)
+                       AND NOT(%s) 
+                       AND (%s)
+                       AND (%I<date_trunc('minute', now())));
+                       $f$,
+                       _rt.recurtype_table, _r.recur_parent_id,
+                       REGEXP_REPLACE(_rt.recurtype_table, E'.*\\.', ''),
+                       REGEXP_REPLACE(_rt.recurtype_table, E'.*\\.', ''),
+                       _rt.recurtype_donecheck,
+                       COALESCE(_rt.recurtype_limit, 'TRUE'),
```
gilmoskowitz: There's a lot of duplicated effort == redundant code separating the schema from the simple table name and double-checking the `recurtype_limit`. Again, this is a case of inheriting from the prior code vs. an opportunity to simplify.

foundation-database/public/functions/createrecurringitems.sql
```diff
@@ -129,6 +198,25 @@ BEGIN
...
+                       REGEXP_REPLACE(_rt.recurtype_table, E'.*\\.', ''),
+                       _rt.recurtype_table,
+                       _r.recur_parent_id,
+                       REGEXP_REPLACE(_rt.recurtype_table, E'.*\\.', ''),
+                       REGEXP_REPLACE(_rt.recurtype_table, E'.*\\.', ''),
+                       _rt.recurtype_donecheck,
+                       COALESCE(_rt.recurtype_limit, 'TRUE'),
+                       _rt.recurtype_schedcol);
+
+    EXECUTE _delstmt;
```
gilmoskowitz: These queries are both built and executed inside a loop, some inside a nested loop. What would it take to parameterize the queries so we can build the strings a few times as possible and simply call them as often as necessary?

foundation-database/public/functions/createrecurringitems.sql
ChristopherCotnoir: That is what the earlier update is for, yes.

PR #2834, opened 2016-09-26
issue #28544 Save without a ship to if no ship to is selected on the Simple Sales …

ChristopherCotnoir:
…Order

### DIFF
--- a/foundation-database/public/tables/metasql/salesOrder-simple.mql
+++ b/foundation-database/public/tables/metasql/salesOrder-simple.mql
@@ -173,12 +173,7 @@
     shipto_shipzone_id
   FROM custinfo LEFT OUTER JOIN cntct billcntct ON (billcntct.cntct_id=cust_cntct_id)
                 LEFT OUTER JOIN addr billaddr ON (billaddr.addr_id=billcntct.cntct_addr_id)
-                LEFT OUTER JOIN shiptoinfo
-                              <? if exists("shipto_id") ?>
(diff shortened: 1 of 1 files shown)


lokeshsh:
Issue is not observed on QA databases

foundation-database/public/tables/metasql/salesOrder-simple.mql
```diff
@@ -173,12 +173,7 @@
     shipto_shipzone_id
   FROM custinfo LEFT OUTER JOIN cntct billcntct ON (billcntct.cntct_id=cust_cntct_id)
                 LEFT OUTER JOIN addr billaddr ON (billaddr.addr_id=billcntct.cntct_addr_id)
-                LEFT OUTER JOIN shiptoinfo
-                              <? if exists("shipto_id") ?>
-                                ON (shipto_id=<? value("shipto_id") ?>)
-                              <? else ?>
-                                ON (shipto_cust_id=cust_id AND shipto_default)
-                              <? endif ?>
+                LEFT OUTER JOIN shiptoinfo ON (shipto_id=<? value("shipto_id") ?>)
```
gilmoskowitz: The old code looks correct - it populates the simple sales order window with appropriate data. Issue 28544 is either a problem when saving the simple sales order (are the right data put into the database?) or when opening the order for edit in the regular sales order window (the previously-saved data are retrieved but the values displayed on the window don't match).

PR #2841, opened 2016-10-03
issue #28634 Prevent all inventory transactions from going negative if disallow ne…

ChristopherCotnoir:
…gative inventory is set.

### DIFF
--- a/foundation-database/public/functions/postinvtrans.sql
+++ b/foundation-database/public/functions/postinvtrans.sql
@@ -112,7 +112,7 @@ BEGIN
 
   IF((_r.itemsite_qtyonhand + round(_sense * pQty, 6)) < 0) THEN
     IF(fetchMetricBool('DisallowNegativeInventory')) THEN
-      RAISE EXCEPTION 'This transaction will cause an item to go negative and negative inventory is currently disallowed [xtuple: postinvtrans, -1]';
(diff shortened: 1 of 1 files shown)


foundation-database/public/functions/postinvtrans.sql
```diff
@@ -110,9 +110,13 @@ BEGIN
     _sense := 1;
   END IF;
 
-  IF((_r.itemsite_costmethod='A') AND (_r.itemsite_qtyonhand + round(_sense * pQty, 6)) < 0) THEN
-    -- Can not let average costed itemsites go negative
-    RAISE EXCEPTION 'This transaction will cause an Average Costed item to go negative which is not allowed [xtuple: postinvtrans, -2]';
+  IF((_r.itemsite_qtyonhand + round(_sense * pQty, 6)) < 0) THEN
+    IF(fetchMetricBool('DisallowNegativeInventory')) THEN
+      RAISE EXCEPTION 'This transaction will cause an item to go negative and negative inventory is currently disallowed [xtuple: postinvtrans, -1]';
```
davidbeauchamp: It appears return code -1 is already used here: https://github.com/ChristopherCotnoir/xtuple/blob/cc8d57b6e35955a2cf6054b2bd41844ac5b20be9/foundation-database/public/functions/postinvtrans.sql#L52 not sure the implications whatever calls this function.

PR #2851, opened 2016-10-10
issue #28697 Don't query lsdetail unless LotSerialControl is set

ChristopherCotnoir:


sindhupriyaav:

PR #2854, opened 2016-10-13
issue #28714 Treat multiple whitespace characters the same as a single whitespace …

ChristopherCotnoir:
…character

### DIFF
--- a/lib/orm/source/xt/pegs/metasql.peg
+++ b/lib/orm/source/xt/pegs/metasql.peg
@@ -242,7 +242,7 @@ Statement
     }
 
 MQLValue
-  = MQLOpen WhiteSpace? Value WhiteSpace? '(' WhiteSpace? id:QuotedIdentifier WhiteSpace? ')' WhiteSpace? MQLClose {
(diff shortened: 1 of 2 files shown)


sindhupriyaav:
Unable to perform UAT. Please provide more information.

lib/orm/source/xt/pegs/metasql.peg
```diff
@@ -242,7 +242,7 @@ Statement
     }
 
 MQLValue
-  = MQLOpen WhiteSpace? Value WhiteSpace? '(' WhiteSpace? id:QuotedIdentifier WhiteSpace? ')' WhiteSpace? MQLClose {
+  = MQLOpen WhiteSpace* Value WhiteSpace* '(' WhiteSpace* id:QuotedIdentifier WhiteSpace* ')' WhiteSpace* MQLClose {
```
gilmoskowitz: A much simpler fix would be to change the definition of `WhiteSpace` to be one or more WS characters.

gilmoskowitz:
@sindhupriyaav - This is a change to a function that is only used internally for now. Only power-users and 3rd party developers would see it. The original bug description shows two queries that should give the same output but do not. The only difference between the two is the number of space characters in the text.

PR #2855, opened 2016-10-14
issue #28719 Add Item Description column

ChristopherCotnoir:


ChristopherCotnoir:
Associated with xtuple/qt-client#1414.

PR #2857, opened 2016-10-17
Allow xt.add_column to modify type and column constraints of existing…

ChristopherCotnoir:
… columns

### DIFF
--- a/lib/orm/source/xt/functions/add_column.sql
+++ b/lib/orm/source/xt/functions/add_column.sql
@@ -20,12 +20,34 @@ begin
   get diagnostics count = row_count;
 
   if (count > 0) then
-    return false;
+    query = 'alter table ' || schema_name || '.' || table_name || ' alter column ' || column_name || ' set data type ' || type_name;
+
+    execute query;
+    if (substring(lower(coalesce(constraint_text, '')) from 'not null') is not null) then
+      query = 'alter table ' || schema_name || '.' || table_name || ' alter column ' || column_name || ' set not null';
+
+      execute query;
+      constraint_text = regexp_replace(constraint_text, 'not null', '', 'i');
+    else
+      query = 'alter table ' || schema_name || '.' || table_name || ' alter column ' || column_name || ' drop not null';
+
+      execute query;
+      constraint_text = regexp_replace(constraint_text, 'null', '', 'i');
+    end if;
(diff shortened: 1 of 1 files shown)


lib/orm/source/xt/functions/add_column.sql
```diff
@@ -20,12 +20,34 @@ begin
   get diagnostics count = row_count;
 
   if (count > 0) then
-    return false;
-  end if;
+    query = 'alter table ' || schema_name || '.' || table_name || ' alter column ' || column_name || ' set data type ' || type_name;
```
gilmoskowitz: Prefer `format()` to string concatenation.

lib/orm/source/xt/functions/add_column.sql
```diff
@@ -20,12 +20,34 @@ begin
...
+
+    execute query;
+    if (substring(lower(coalesce(constraint_text, '')) from 'not null') is not null) then
+      query = 'alter table ' || schema_name || '.' || table_name || ' alter column ' || column_name || ' set not null';
```
gilmoskowitz: These constraint changes may cause table rewrites. Therefore we need to test performance with a good-sized database before committing to this step-by-step, just-do-it strategy.

lib/orm/source/xt/functions/add_column.sql
```diff
@@ -20,12 +20,34 @@ begin
...
+      constraint_text = regexp_replace(constraint_text, 'not null', '', 'i');
+    else
+      query = 'alter table ' || schema_name || '.' || table_name || ' alter column ' || column_name || ' drop not null';
 
-  query = 'alter table ' || schema_name || '.' || table_name || ' add column ' || column_name || ' ' || type_name || ' ' || coalesce(constraint_text, '');
+      execute query;
+      constraint_text = regexp_replace(constraint_text, 'null', '', 'i');
+    end if;
+    if (substring(lower(coalesce(constraint_text, '')) from 'default') is not null) then
+      query = 'alter table ' || schema_name || '.' || table_name || ' alter column ' || column_name || ' set default ' || regexp_replace(constraint_text, 'default', '', 'i');
```
gilmoskowitz: One specific consideration involves setting defaults. The [PostgreSQL docs](https://www.postgresql.org/docs/current/static/ddl-alter.html) include this note: "Tip: Adding a column with a default requires updating each row of the table (to store the new column value)."

PR #2877, opened 2016-11-01
issue #28820 fix for createwf function

bixwilson:


sindhupriyaav:
Performed UAT: Issue still exists. Purchase activities not created when PO is created.

rkannadevara:
Performed UAT: Workflow activities are not displayed on mobile web client for a purchase order when workflow is created from desktop application.

bixwilson:
Purchase order activities are only created when the PO is released. I just tested and confirmed that the PO workflow activities are created and display correctly in the desktop and the web client, after the PO is released.

PR #2880, opened 2016-11-03
Fix XM.SalesOrder workflow test for pri-ext with metric 'TriggerWorkflow' on.

bendiy:
Will have a merge conflict with https://github.com/xtuple/xtuple/pull/2877 after that's been merged.

### DIFF
--- a/database/source/xt/functions/createwf.sql
+++ b/database/source/xt/functions/createwf.sql
@@ -4,35 +4,63 @@ $BODY$
 /* Copyright (c) 1999-2016 by OpenMFG LLC, d/b/a xTuple.
    See www.xm.ple.com/CPAL for the full text of the software license. */
 DECLARE
-tg_table_name ALIAS FOR $1; 
+tg_table_name ALIAS FOR $1;
 tg_table_row  ALIAS FOR $2;
 _wftypecode   TEXT      := '';
 _source_model TEXT      := '';
 _workflow_class TEXT    := '';
+_default_hold_type CHARACTER(1);
 _item_uuid    UUID;
 _parent_id    INTEGER   := 0;
 _order_id     INTEGER   := 0;
 _poparent     RECORD;
+_saletypeextExists BOOLEAN := false;
 
 BEGIN
 
    IF (fetchmetricbool('TriggerWorkflow')) THEN
-            
-      -- Get data by type      
+
+      -- Get data by type
       IF (tg_table_name = 'cohead') THEN
-         _wftypecode := 'SO';
-         _workflow_class  := 'XM.SalesOrderWorkflow';
-         _item_uuid := tg_table_row.obj_uuid;
-         _parent_id := tg_table_row.cohead_saletype_id;
-         _order_id := tg_table_row.cohead_id;
-         
+        _wftypecode := 'SO';
+        _workflow_class  := 'XM.SalesOrderWorkflow';
+        _item_uuid := tg_table_row.obj_uuid;
+        _parent_id := tg_table_row.cohead_saletype_id;
+        _order_id := tg_table_row.cohead_id;
+
+        -- Copy SaleType char to Sales Order.
+        PERFORM xt.copychars('ST', _parent_id, 'S', _order_id);
+
+        -- Switch `cohead_holdtype` to the `saletypeext_default_hold_type`.
+        SELECT EXISTS (
+          SELECT 1
+          FROM   pg_catalog.pg_class c
+          JOIN   pg_catalog.pg_namespace n ON n.oid = c.relnamespace
+          WHERE  n.nspname = 'xt'
+          AND    c.relname = 'saletypeext'
+          AND    c.relkind = 'r' -- table
+        ) INTO _saletypeextExists;
+
+        IF (_saletypeextExists) THEN
+          SELECT
+            saletypeext_default_hold_type INTO _default_hold_type
+          FROM xt.saletypeext
+          WHERE saletypeext_id = _parent_id;
+
+          IF ((_default_hold_type IS NOT NULL) AND (_order_id > 0)) THEN
+            UPDATE cohead SET
+              cohead_holdtype = _default_hold_type
+            WHERE cohead_id = _order_id;
+          END IF;
+        END IF;
+
       ELSIF (tg_table_name = 'prj') THEN
          _wftypecode := 'PRJ';
          _workflow_class  := 'XM.ProjectWorkflow';
          _item_uuid := tg_table_row.obj_uuid;
          _parent_id := tg_table_row.prj_prjtype_id;
          _order_id := tg_table_row.prj_id;
-         
+
       ELSIF (tg_table_name = 'poheadext') THEN
          _wftypecode := 'PO';
          _workflow_class  := 'XM.PurchaseOrderWorkflow';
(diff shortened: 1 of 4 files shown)


bendiy:
@bixwilson @gilmoskowitz Let me know when https://github.com/xtuple/xtuple/pull/2877 gets merged and I can address the merge conflict.

database/source/xt/functions/createwf.sql
```diff
@@ -4,35 +4,63 @@ $BODY$
...
+          AND    c.relkind = 'r' -- table
+        ) INTO _saletypeextExists;
+
+        IF (_saletypeextExists) THEN
+          SELECT
+            saletypeext_default_hold_type INTO _default_hold_type
+          FROM xt.saletypeext
+          WHERE saletypeext_id = _parent_id;
+
+          IF ((_default_hold_type IS NOT NULL) AND (_order_id > 0)) THEN
```
gilmoskowitz: 1. What is the rationale for not updating the hold type if the default `IS NOT NULL`? If this is important, the reason is not apparent. It should be.
2. The inner `IF` and `_default_hold_type` are also unnecessary. Simplify:
```
if (the saletypeext table exists and we have an order to process)
  set the cohead_holdtype to the default
```
```plpgsql
if _order_id > 0 and exists(select 1 from pg_class ...) then
  update cohead set cohead_holdtype = salestypeext_default_hold_type
    from saletypeext
     ...
end if
```
Is there a circumstance under which we can create workflow without a saletypeext table? If not then we're working much too hard to solve the problem.

database/source/xt/functions/createwf.sql
```diff
@@ -67,14 +95,14 @@ BEGIN
...
+
       -- Get _source_model
       SELECT wftype_src_tblname AS srctblname FROM xt.wftype WHERE wftype_code = _wftypecode INTO _source_model;
       IF (NOT FOUND) THEN
         RAISE WARNING 'Missing sourceModel needed to generate workflow!';
       END IF;
-     
-      -- Generate workflow 
+
+      -- Generate workflow
```
gilmoskowitz: Several of these comments don't seem to add value. Remove 'em?

lib/backbone-x/source/characteristic.js
```diff
@@ -195,7 +195,9 @@ white:true*/
     },
 
     characteristicDidChange: function (model, value, options) {
-      this.set('value', '');
+      if (!model.get("value")) {
+        this.set('value', '');
```
gilmoskowitz: This reads as `if model.value is falsey then set this.value to falsey`. The rationale for bothering with the code change is unclear. Are we just trying to normalize falsey values?

test/specs/sales_order.js
```diff
@@ -368,86 +372,201 @@ setTimeout:true, before:true, clearTimeout:true, exports:true, it:true, describe
...
-        salesOrderModel.set({saleType: null});
-        salesOrderModel.get("workflow").reset([]);
-        salesOrderModel.get("characteristics").reset([]);
-      });
+
+        /**
+          @member -
+          @memberof SalesOrder
+          @description When the sale type changes, the characteristics of the new sale type
+            are copied over to the sales order.
```
gilmoskowitz: A lot of these documentation comments in the test are wasted. They don't tell us anything we don't already know from the `it()` and the filename.

test/specs/sales_order.js
```diff
@@ -368,86 +372,201 @@ setTimeout:true, before:true, clearTimeout:true, exports:true, it:true, describe
...
+          var copiedWorkflow;
+
+          saleTypeModel.get("workflow").models[0]
+            .set({workflowType: XM.SalesOrderWorkflow.TYPE_PACK});
+          salesOrderModel.set({packDate: new Date("1/1/2004")});
+          salesOrderModel.set({saleType: saleTypeModel});
+          assert.equal(salesOrderModel.get("workflow").length, 1);
+          copiedWorkflow = salesOrderModel.get("workflow").models[0];
+          assert.equal(copiedWorkflow.get("dueDate").getDate(), new Date("1/1/2004").getDate());
+        });
```
gilmoskowitz: If we know how to test this, why `it.skip()`? If we don't know how to test it or don't have the resources to test it (e.g. th `inventory` extension), why include the code for it? If it's going to be reimplemented in inventory, what's going to be different there?

test/specs/sales_order.js
```diff
@@ -368,86 +372,201 @@ setTimeout:true, before:true, clearTimeout:true, exports:true, it:true, describe
...
+
+        /**
+          @member -
+          @memberof SalesOrder
+          @description The due date for "Ship" workflow items will default to the schedule
+            date on the header. If that date changes, "Ship" workflow items will be updated.
+        */
+        // TODO: reimplement in inventory
+        it.skip("The due date for Ship workflow items will default to the schedule date on the order",
```
gilmoskowitz: ditto:
- why `it.skip()`?
- why include the code?
- what's going to be different in `inventory`?

lib/backbone-x/source/characteristic.js
bendiy: This is a bug in the code I found while fixing this issue. The `characteristicDidChange()` method is called when you add a char to the `XM.SalesOrder.characteristics[]` array. The characteristics array is made up of `XM.SalesOrderCharacteristic` objects. They have five properties `id`, `uuid`, `salesOrder`, `characteristic` and `value`. Same goes for other types of chars. The `characteristic` property is basically `charass_char_id` and the `value` property is `charass_value`. You can add a `charass` record without setting a value. The old code here always defaults the `value` to an empty string.

However, it's possible to pass in a `XM.SalesOrderCharacteristic` object as the `model` argument that already _has_ a `value` set. This code was incorrectly switching that `value` to an empty string.

It is possible to pass in a `XM.SalesOrderCharacteristic` object on the `model` argument without any `value` property. This code was intended to initialized it to an empty string so `value` _did_ exists, but it was always doing it vs. only when not set.

test/specs/sales_order.js
bendiy: ditto
Moving old code to maintain functional scope.

test/specs/sales_order.js
bendiy: Agreed. I can clean up this old code as I found these documentation comments to be useless.

database/source/xt/functions/createwf.sql
```diff
@@ -4,35 +4,63 @@ $BODY$
...
 tg_table_row  ALIAS FOR $2;
 _wftypecode   TEXT      := '';
 _source_model TEXT      := '';
 _workflow_class TEXT    := '';
+_default_hold_type CHARACTER(1);
 _item_uuid    UUID;
 _parent_id    INTEGER   := 0;
 _order_id     INTEGER   := 0;
 _poparent     RECORD;
+_saletypeextExists BOOLEAN := false;
```
bendiy: It's for readability. I find putting an entire seven line query inside an `IF (...)` hard to scan.

database/source/xt/functions/createwf.sql
bendiy: 1. AFAIK, `cohead_holdtype` is never `NULL`, it gets set to `'N'` for "no hold". If the `saletypeext_default_hold_type` was not specified when defining that Sale Type's settings, then I didn't want to change the Sales Order's hold type to `NULL` or even `'N'`. It should stay as it was because the Sales Type isn't configured to set a hold.

2. I'll simplify these conditionals a bit.

> Is there a circumstance under which we can create workflow without a saletypeext table? If not then we're working much too hard to solve the problem.

I do not know the correct answer to that. Currently, `xt.saletypeext` is webenabled only. We are trying to get workflow into core. If a non-webenabled database calls this, the hold type copying feature will be skipped. We should consider adding `default_hold_type` to the core `saletype` table, but that's out of scope for this CI test failure fix. There are several other features of workflow documented in the tests that are not in core yet.

gilmoskowitz:
@bendiy - Please fix the merge conflicts and remove useless stuff. If you think the `.skip`ped tests have useful content, leave 'em. Also add a 1-line comment explaining the hold type handling; if it takes more than one line then either there's too much detail or something needs to be clarified somewhere else.

bendiy:
@gilmoskowitz Updated.

PR #2884, opened 2016-11-06
Issue #25077: resolve margin calc on Invoice items list

andersdd:


sindhupriyaav:
Performed UAT: Issue is fixed

gilmoskowitz:
Note: I don't see way to fix a UOM conversion failure if there is no UOM conversion ratio.

PR #2887, opened 2016-11-08
Set null values to default when changing a column from nullable witho…

ChristopherCotnoir:
…ut default to not null with a default

### DIFF
--- a/lib/orm/source/xt/functions/add_column.sql
+++ b/lib/orm/source/xt/functions/add_column.sql
@@ -25,6 +25,7 @@ begin
 
   if (count > 0) then
     if(_current.type !~~* type_name and type_name !~* 'serial') then
+--The operator !~~* when used with no special regular expression operators amounts to a straight case-insensitive comparison
       query = format('alter table %I.%I alter column %I set data type %s', schema_name, table_name, column_name, type_name);
 
       execute query;
@@ -48,6 +49,9 @@ begin
       constraint_default = trim(regexp_replace(coalesce(constraint_text, ''), 'null', '', 'i'));
     end if;
 
+--The above does not behave correctly in most cases wherein a constraint or the default value
+--contains the text 'not null' or 'null' (e.g. constraint_text='default ''not null''')
+
     if (constraint_default ~* 'default') then
       if (constraint_default ~* 'check') then
         if(strpos(constraint_default, 'default') < strpos(constraint_default, 'check')) then
@@ -56,6 +60,12 @@ begin
           constraint_default = trim((regexp_matches(constraint_default, 'check.*(default .*)', 'i'))[1]::text);
         end if;
       end if;
+
+--The above only handles check constraints, and only if the default value does not contain
+--the test 'check' and the check constraint does not contain the text 'default'.
+--It does not handle unique constraints, primary keys (though these should never be used with a default),
+--foreign key constraints, or exclude constraints correctly.
+--Check constraints are also simply ignored on existing columns rather than added/updated idempotently
 
       constraint_default = trim(regexp_replace(constraint_default, 'default', '', 'i'));
       if(coalesce(_current.defaultval, '')!=constraint_default) then
--- a/lib/orm/source/xt/functions/add_column.sql
+++ b/lib/orm/source/xt/functions/add_column.sql
@@ -4,6 +4,9 @@ create or replace function xt.add_column(table_name text, column_name text, type
 declare
   count integer;
   query text;
+  constraint_null boolean;
+  constraint_default text;
+  column_null boolean := false;
   comment_query text;
   _current record;
 begin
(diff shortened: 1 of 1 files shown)


lib/orm/source/xt/functions/add_column.sql
```diff
@@ -21,38 +24,61 @@ begin
...
+
+    if (constraint_default ~* 'default') then
+      constraint_default = trim(regexp_replace(constraint_default, 'default', '', 'i'));
+      if(coalesce(_current.defaultval, '')!=constraint_default) then
+        query = format('alter table %I.%I alter column %I set default %s', schema_name, table_name, column_name, constraint_default);
 
         execute query;
+
+        if(_current.defaultval is null and constraint_null and column_null) then
+          query = format('update %I.%I set %I=%s where %I is null', schema_name, table_name, column_name, constraint_default, column_name);
```
bendiy: `set %I=%s` and `constraint_default`.
> `s` formats the argument value as a simple string. A null value is treated as an empty string.

https://www.postgresql.org/docs/9.3/static/functions-string.html#FUNCTIONS-STRING-FORMAT

What if `constraint_default` is a function call, not a string? e.g. `uuid_generate_v4()` or `defaultlocale()` etc.

bendiy:
This change is probably going to surface a few issues:
```
  createpkgschema()
    ◦ needs no existing pkg to test properly:   ✓ needs no existing pkg to test properly 
    ◦ should create a schema: <<ERROR 2016-11-08T21:44:03.636Z>> Unrecoverable exception. expected [error: null value in column "pkghead_id" violates not-null constraint] to equal nullexpected [error: null value in column "pkghead_id" violates not-null constraint] to equal null
```

bendiy:
https://github.com/xtuple/xtuple/blob/4_11_x/test/database/functions/createpkgschema.js#L25
https://github.com/xtuple/xtuple/blob/4_11_x/foundation-database/public/tables/pkghead.sql#L3

bendiy:
@ChristopherCotnoir Have you attempted to log a build's call to this function?
I want to make sure we catch and handle all use cases. I came across another today:
```
select xt.add_column('item','item_web_published', 'boolean', 'NOT NULL DEFAULT FALSE CHECK (item_active AND item_sold)', 'public', 'The Item is published on the web site.');
```
`constraint_text` can have a `CHECK` constraint in it.

ChristopherCotnoir:
I haven't attempted to log a build. I'm not entirely sure how to go about that, or how I could meaningfully sift through all the data.

bendiy:
If you do a `RAISE WARNING` and log the `constraint_text` along with a unique string you can grep for (e.g. `##@@find-me@@## add_column() constraint_text = 'foo'`), it should be possible to pull out only the calls to this function in the postgres log. Or insert `constraint_text` into a throw away table you add so it can be sorted and parsed after a build.

I am looking to do a little analytics to make sure we catch all the edge cases.

I played around with the `CHECK` constraints a little more yesterday and I don't think adding them through `add_column()` is the best way to do it. You don't have control over naming the constraint. They just become `tablename_check`, `tablename_check_1`, `tablename_check_n`. That means we cannot update or remove them in the future. I moved the ones I was using to an `xt.add_constraint()` call, but I doubt this was the only place that had/has `CHECK` constraints in `add_column()`. The `xt.add_constraint()` function has mainly been used for adding foreign key constraints. It works perfectly find for `CHECK` constraints too, but I bet some of those have use `xt.add_column()` and we really should fix those.

I propose we add an assert in the `add_column()` function that `constraint_text` cannot include any `CHECK` constraints and `RAISE EXCEPTION 'Do not use xt.add_column() to add constraints. Its use for that has been depreciated. Use xt.add_constraint() instead.'` if any are found.

gilmoskowitz:
@bendiy - Forbidding creating specific kinds of constraint with `xt.add_column()` because they're somehow inconvenient is bad behavior. We should accept whatever the caller gives us. We might choose not to take advantage of the feature but that's different from disallowing it.

lib/orm/source/xt/functions/add_column.sql
```diff
@@ -21,38 +24,61 @@ begin
   get diagnostics count = row_count;
 
   if (count > 0) then
-    if(lower(_current.type)!=lower(type_name) and type_name !~* 'serial') then
+    if(_current.type !~~* type_name and type_name !~* 'serial') then
```
gilmoskowitz: @ChristopherCotnoir - `!~~*` still needs to be fixed

lib/orm/source/xt/functions/add_column.sql
```diff
@@ -57,6 +61,12 @@ begin
         end if;
       end if;
 
+--The above only handles check constraints, and only if the default value does not contain
+--the test 'check' and the check constraint does not contain the text 'default'.
```
bendiy: `the test 'check'` -> `the text 'check'`

PR #2889, opened 2016-11-09
issue #28877 Don't fetch item sources without prices on screens that are listing t…

ChristopherCotnoir:
…he prices for a specific item source


ChristopherCotnoir:
Associated with xtuple/qt-client#1446.

PR #2893, opened 2016-11-15
Issue #28170: Bank Rec add foreign key to prevent orphaned records

andersdd:
Seen this happen in a few Dbs


lokeshsh:
Performed UAT: Observed that code is added to the 'bankrecitem.sql' file

PR #2894, opened 2016-11-15
backport workflow changes from 4_11_x

gilmoskowitz:
cherry-pick 080639c fc856b3 41a4472
requires https://github.com/xtuple/private-extensions/pull/865


database/source/xt/functions/createwf.sql
```diff
@@ -1,4 +1,8 @@
-CREATE OR REPLACE FUNCTION xt.createwf(text, anyelement)
+DROP FUNCTION IF EXISTS xt.createwf(text, anyelement);
```
bendiy: I think we need to move this function to `commercialcore/foundation-database/workflow/sql/` so it's in the commercial non-webenabled build.

PR #2895, opened 2016-11-15
issue #16967 Properly handle recurtype tables without schema and with a schedcol t…

ChristopherCotnoir:
…hat is a date and not a timestamp


rkannadevara:
Performed UAT: Issue is fixed. But observed the following behavior for recurring vouchers:
1. Enabled 'Recurring' in 'Advanced' tab in 'Miscellaneous Voucher' screen'
2. Create no more than' is selected as '5' for 'Every' 1 day.
3. On saving the voucher, only one voucher is created and recurring vouchers are not created.

Also observed that recurring invoices are created as expected.

sindhupriyaav:

PR #2897, opened 2016-11-15
Issue #27791: Work Order and Rental Item are grayed out on "New" menu of documents tab

davidbeauchamp:
Issue is two part, with Work Order there was a typo in the privilege name, work Rental Item the "create" privilege column should be empty as those do not need to be arbitrarily created from the documents tab.


davidbeauchamp:
Requires https://github.com/xtuple/private-extensions/pull/866

rkannadevara:
Performed UAT: Observed that 'Rental' item is not displayed in the list under 'New' as expected. But on clicking 'Work Order' under 'New', observed the below error message:
wodocument

PR #2899, opened 2016-11-18
Issue #25077: Revise Invoice Item margin to handle different Price Uoms

andersdd:
Tested with Sales Order invoices and Misc. Invoices.   With and without Price UOMs


lokeshsh:
Performed UAT: Issue is fixed

PR #2903, opened 2016-11-18
issue #28924 Don't query xtmfg tables or functions without the package

ChristopherCotnoir:


lokeshsh:
Performed UAT: Issue is fixed

PR #2904, opened 2016-11-21
Issue #28882: Add tax field to api.purchaseline

andersdd:
Not adding tax_recoverable field as this is not available on the ui and is not implemented as yet (may never be implemented?)


lokeshsh:
Performed UAT: 'tax_type_next' column has been added to 'api.purchaseline' and tax type value is being displayed for the purchase order.

PR #2908, opened 2016-11-27
Issue #28564: Remove non-physical items (item types) from inventory reports

andersdd:


rkannadevara:

PR #2910, opened 2016-11-29
Issue #27705: Resolve Item characteristic propagation to order documents

andersdd:
Only propagate to order item chars if selected in characteristic setup.


gilmoskowitz:
👍
@rkshapiro - As long as you are confident it fixes the bug, keep going with https://github.com/xtuple/qt-client/wiki/Bug-Fix-Workflow#completion---reviewer. There are ways to simplify and clarify the code even more; it's up to you to decide whether that effort is warranted, as the new version is already much simpler than the previous version.

PR #2915, opened 2016-12-01
Issue #28799: Add Work Order close date for stats and dashboarding

andersdd:


rkannadevara:
Performed UAT: Observed that Work order close date is not observed on the dashboards. Please provide more information to verify the fix.

andersdd:
@rkannadevara This change was to collect data so it was _available_ for proposed dashboards, not develop the dashboards themselves.  Just check that the new column is added to the wo table and populated when the WO is closed.

lokeshsh:
Performed UAT: Observed that a column 'wo_closedate time stamp with time zone' has been added to WO table.

PR #2916, opened 2016-12-07
issue #28637 Properly convert currency

ChristopherCotnoir:


sindhupriyaav:
Unable to verify the fix . Please provide more information.

ChristopherCotnoir:
Parts 1 and 2. 3 and 4 are correct behavior, they were just inconsistent with the aging report because the aging report was wrong.

PR #2917, opened 2016-12-08
issue #29005 Ensure sequences are given proper permissions

ChristopherCotnoir:


sindhupriyaav:
Performed UAT by following the below mentioned steps:

1. Created Quality test plan and Quality test for a 'YPAINT1'(Lot controlled) product 
2. Assigned Quality and All privilleges to Usge User
3. As a Usge user created Work order, released and done Post Production  without any error.

Please provide more information if above followed steps are not correct.

lib/orm/source/xt/functions/add_column.sql
```diff
@@ -1,118 +1,129 @@
...
+      END IF;
+    END IF;
+  ELSE
+    _query := format('ALTER TABLE %I.%I ADD COLUMN %I %s %s', pSchema, pTable, pColumn, pType, COALESCE(pConstraint, ''));
+
+    EXECUTE _query;
+  END IF;
+
+  IF (pType ~~* 'serial') THEN
+    _sequence := format('%I_%I_id_seq', pTable, pTable);
```
gilmoskowitz: This ` format()` call does not work as expected. If `pTable` isn't all lowercase then the sequence name will wind up with `"` embedded in it. As a result, the following `if exists()` will fail.

PR #2923, opened 2016-12-15
issue #28597 Attempt to resolve rounding issues when applying credit memos to vouc…

ChristopherCotnoir:
…hers in different currency


ChristopherCotnoir:
Requires xtuple/qt-client#1494.

PR #2925, opened 2016-12-21
issue #29133 - debugging changes for issue #28896 were too aggressive

gilmoskowitz:
Bug 29133 was caused by incorrectly converting warnings to exceptions when tracking down bug 28896. This was based on the false assumption that certain missing data ("parent" records) indicated a bug earlier in the processing. In fact those data trigger workflow generation if present but missing data should be ignored.

### DIFF
--- a/database/source/xt/functions/createwf.sql
+++ b/database/source/xt/functions/createwf.sql
@@ -59,7 +59,7 @@ BEGIN
         FROM pohead JOIN xt.poheadext ON poheadext_id = pohead_id
        WHERE pohead_id = tg_table_row.pohead_id;
       IF _parent_id IS NULL THEN
-        RAISE EXCEPTION 'Cannot find parentId needed to generate workflow [xtuple: createwf, -1, %, %]',
+        RAISE DEBUG 'Cannot find parentId needed to generate workflow [%, %]',
                         tg_table_name, tg_table_row.pohead_id;
       END IF;
       IF _poStatus <> 'O' THEN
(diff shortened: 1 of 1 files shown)


bixwilson:
@gilmoskowitz That doesn't seem right. I don't know of a case where the createwf function would be called but the data should be ignored.

gilmoskowitz:
https://github.com/xtuple/private-extensions/blob/4_11_x/source/commercialcore/foundation-database/workflow/sql/createwf_triggers.sql generates indiscriminate calls to `createwf()`. Any update to the tables monitored for workflow activity could raise errors if the workflow system is installed but not configured (i.e. the triggers exist). The fields that drive workflow are not all required (e.g. `pohead_potype_id`).

bixwilson:
Ok. I guess this problem exists in part because the parts of workflow are still not all in core. Thanks.

sindhupriyaav:
Performed UAT: Able to release purchase order in all data bases( Enterprise, Distribution, Distribution empty). However, 'Error Retrieving RA Information' dialog with db log error is displayed while creating new Return Authorization in Enterprise and Distribution editions. Please refer to DB log error below.

ERROR:  Improper table supplied to createwf function [xtuple: createwf, -2, rahead]
CONTEXT:  SQL statement "SELECT xt.createwf(TG_TABLE_NAME, NEW)"
PL/pgSQL function createwf_after_insert() line 5 at PERFORM
(P0001) QPSQL: Unable to create query

The Query:
INSERT INTO rahead (    rahead_id, rahead_number, rahead_authdate) VALUES (    :rahead_id, COALESCE(:rahead_number,0), :rahead_authdate);

Bound Values:
:rahead_authdate:	
:rahead_id:	14
:rahead_number:	80011

gilmoskowitz:
https://github.com/xtuple/private-extensions/pull/879 fixes the rahead problem found during testing.

rkannadevara:

PR #2937, opened 2017-01-05
Issue #29173: Resolve erroneous sort order in GL Transactions report 

andersdd:
When Opening balances are selected


sindhupriyaav:

PR #2941, opened 2017-01-08
Issue #29216: Resolve sql bug in copyproject function

andersdd:


lokeshsh:

PR #2947, opened 2017-01-16
Issue #29247: Revert trigger function to prevent duplicated SO characteristics

andersdd:
This reverts the change made as part of PR #2644 as that was causing the duplicated characteristics.

Need to check this change with @bendiy and @rkshapiro as they need to provide the sales order line with the `coitem_imported` flag set to `true`.

### DIFF
--- a/foundation-database/public/trigger_functions/coitem.sql
+++ b/foundation-database/public/trigger_functions/coitem.sql
@@ -1,5 +1,5 @@
-﻿CREATE OR REPLACE FUNCTION _soitemTrigger() RETURNS TRIGGER AS $$
--- Copyright (c) 1999-2016 by OpenMFG LLC, d/b/a xTuple.
+CREATE OR REPLACE FUNCTION _soitemTrigger() RETURNS TRIGGER AS $$
+-- Copyright (c) 1999-2017 by OpenMFG LLC, d/b/a xTuple.
 -- See www.xtuple.com/CPAL for the full text of the software license.
 DECLARE
   _changelog BOOLEAN := FALSE;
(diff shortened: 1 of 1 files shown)


rkshapiro:
Ben found that the coitem_imported is not exposed to the rest api. Since xtuplecommerce does set cohead_imported to TRUE, he suggested that the trigger check for both coitem_imported and cohead_imported. That should cover the online order case.

bendiy:
I've been dealing with something related to this issue and I think we should set a `BEFORE INSERT ON charass` that does an `UPSERT` if the characteristic for the `coitem` already exists.
This current trigger adds the default chars for the `coitem` on `INSERT`, but if I need the website to set different `coitem` characteristic values than the defaults, I get duplicates because the characteristics passed from the website are inserted later, after these defaults were created by this trigger.

The `coitem` characteristics are unique, you can't add the same characteristic to a line twice in the Qt Client. Doing an `UPSERT` would be consistent with that.

andersdd:
Added check in the trigger for `cohead_imported` as discussed with @rkshapiro 

@bendiy Need to be very careful with coitem characteristics as they are used in configurable items and possibly propagate out to WO materials and WO chars as well.

bendiy:
@andersdd That's my use case. Configured Item that generates a Work Orders. The Qt Client triggers the WO creation in C++. I'm doing it over the API from xTupleCommerce with a `coitem` trigger looking for  `INT` Sales Type.

andersdd:
@bendiy I've changed the logic to look for the imported flag on cohead or coitem.  This only triggers on INSERT of coitems.

PR #2949, opened 2017-01-19
issue #29137 Insert new reports at grade 0 if it is not already taken, and fix rep…

ChristopherCotnoir:
…orts incorrectly inserted at grade 1

### DIFF
--- a/foundation-database/public/functions/saveobject.sql
+++ b/foundation-database/public/functions/saveobject.sql
@@ -1,0 +1,130 @@
+CREATE OR REPLACE FUNCTION saveObject(pType TEXT, pGroup TEXT, pName TEXT, pGrade INTEGER, pSource TEXT, pNotes TEXT, pEnabled BOOLEAN, pSchema TEXT) RETURNS INTEGER AS $$
+-- Copyright (c) 1999-2017 by OpenMFG LLC, d/b/a xTuple. 
+-- See www.xtuple.com/CPAL for the full text of the software license.
+DECLARE
+  _schema TEXT;
+  _table TEXT;
+  _idcolumn TEXT;
+  _namecolumn TEXT;
+  _gradecolumn TEXT;
+  _sourcecolumn TEXT;
+  _notescolumn TEXT;
+  _namecondition TEXT;
+  _groupinsertclause TEXT := '';
+  _groupinsertvalue TEXT := '';
+  _enabledupdateclause TEXT := '';
+  _enabledinsertclause TEXT := '';
+  _enabledinsertvalue TEXT := '';
+  _grade INTEGER;
+  _id INTEGER;
+
+BEGIN
+  pType := lower(pType);
+  pSchema := lower(pSchema);
+
+  IF (pSchema IS NULL) THEN
+    _schema := 'public';
+    _table := pType;
+  ELSE
+    _schema := pSchema;
+    _table := format('pkg%s', pType);
+  END IF;
+
+  IF (pType='script') THEN
+    -- scripts are handled differently because they use order instead of grade
+    -- uiforms use a column named uiform_order, but it is really a grade
+    EXECUTE format($f$
+                      UPDATE ONLY %I.%I
+                      SET script_source=%L,
+                      script_notes=%L,
+                      script_enabled=%L
+                      WHERE script_name=%L
+                      AND script_order=%L
+                      RETURNING script_id;
+                    $f$, _schema, _table, pSource, pNotes, pEnabled, pName, pGrade) INTO _id;
+
+    IF (NOT FOUND) THEN
+      EXECUTE format($f$
+                       INSERT INTO %I.%I
+                       (script_name, script_order, script_source, script_notes, script_enabled)
+                       VALUES(%L, %L, %L, %L, %L)
+                       RETURNING script_id;
+                     $f$, _schema, _table, pName, pGrade, pSource, pNotes, pEnabled) INTO _id;
+    END IF;
+  ELSE
+    IF (pType='metasql') THEN
+      _namecondition := format('metasql_group=%L AND metasql_name=%L', pGroup, pName);
+      _groupinsertclause := 'metasql_group,';
+      _groupinsertvalue := format('%L,', pGroup);
+    ELSE
+      _namecondition := format('%I=%L', format('%s_name', pType), pName);
+    END IF;
+
+    IF (pType='uiform') THEN
+      _gradecolumn := 'uiform_order';
+    ELSE
+      _gradecolumn := format('%s_grade', pType);
+    END IF;
+
+    _idcolumn := format('%s_id', pType);
+
+    IF (pType='uiform') THEN
+      _enabledupdateclause := format(',uiform_enabled=%L', pEnabled);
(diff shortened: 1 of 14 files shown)


ChristopherCotnoir:
Associated with xtuple/updater#48.

rkshapiro:
@gilmoskowitz I am unfamiliar with this section of the system. Can you review this change?

foundation-database/public/functions/saveobject.sql
```diff
@@ -0,0 +1,130 @@
+CREATE OR REPLACE FUNCTION saveObject(pType TEXT, pGroup TEXT, pName TEXT, pGrade INTEGER, pSource TEXT, pNotes TEXT, pEnabled BOOLEAN, pSchema TEXT) RETURNS INTEGER AS $$
```
gilmoskowitz: This function name is too generic. Maybe `saveExtensionObject()`?

foundation-database/public/functions/saveobject.sql
```diff
@@ -0,0 +1,130 @@
...
+    -- scripts are handled differently because they use order instead of grade
+    -- uiforms use a column named uiform_order, but it is really a grade
+    EXECUTE format($f$
+                      UPDATE ONLY %I.%I
+                      SET script_source=%L,
+                      script_notes=%L,
+                      script_enabled=%L
+                      WHERE script_name=%L
+                      AND script_order=%L
+                      RETURNING script_id;
```
gilmoskowitz: Watch indentation here and elsewhere

foundation-database/public/functions/saveobject.sql
```diff
@@ -0,0 +1,130 @@
...
+                    $f$, _schema, _table, pSource, pNotes, pEnabled, pName, pGrade) INTO _id;
```
gilmoskowitz: `COALESCE(pOrder, 0)` and `COALESCE(pEnabled, true)`

foundation-database/public/functions/saveobject.sql
```diff
@@ -0,0 +1,130 @@
...
+    ELSE
+      EXECUTE format($f$
+                       INSERT INTO %I.%I
+                       (%s %I, %I, %I, %I%s)
+                       VALUES(%s %L, %L, %L, %L%s)
+                       RETURNING %I;
+                     $f$, _schema, _table, _groupinsertclause, _namecolumn, _gradecolumn, _sourcecolumn, _notescolumn, _enabledinsertclause, _groupinsertvalue, pName, _grade, pSource, pNotes, _enabledinsertvalue, _idcolumn) INTO _id;
+    END IF;
+  END IF;
+
```
gilmoskowitz: Validate `pType` somewhere. A call with an invalid `pType` is probably a bug.

foundation-database/public/functions/saveobject.sql
```diff
@@ -0,0 +1,130 @@
...
+                     SELECT MIN(sequence_value-1)
+                     FROM sequence
+                     WHERE sequence_value-1>=%L
+                     AND sequence_value-1 NOT IN (SELECT %I
+                                                  FROM %I
+                                                  JOIN pg_class c ON %I.tableoid=c.oid
+                                                  JOIN pg_namespace n ON c.relnamespace=n.oid
+                                                  WHERE %s
+                                                  AND n.nspname!=%L)
+                   $f$, pGrade, _gradecolumn, pType, pType, _namecondition, _schema) INTO _grade;
```
gilmoskowitz: This looks like it's selecting the lowest unused grade in this extension's table or the `public` equivalent. If that's correct, the join with pg_class and pg_namespace can be avoided by `select ... from only %I.%I ...`

foundation-database/public/functions/saveobject.sql
```diff
@@ -0,0 +1,130 @@
...
+
+    IF (_id IS NOT NULL) THEN
+      EXECUTE format($f$
+                       UPDATE %I.%I
+                       SET %I=%L,
+                           %I=%L
+                           %s
+                       WHERE %I=%L;
+                     $f$, _schema, _table, _sourcecolumn, pSource, _notescolumn, pNotes, _enabledupdateclause, _idcolumn, _id);
+    ELSE
```
gilmoskowitz: Could use the same "update, then insert if not found" approach as was used for scripts above.

scripts/lib/util/convert_specialized.js
```diff
@@ -72,84 +64,31 @@ regexp:true, undef:true, strict:true, trailing:true, white:true */
...
-                "    update " + tableName + " set" +
-                "      report_descrip = $description$" + description + "$description$," +
-                "      report_source = $content$" + content + "$content$," +
-                "      report_loaddate = now() " +
-                "     where report_name = '" + name + "'" +
-                "      and report_grade = _grade;" +
-                "  end if;" +
-                " end $do$;";
-    return upsertSql;
+    return "select saveObject('report'," + "NULL,'" + name + "'," + parseInt(grade) +
```
gilmoskowitz: Don't `+` two constants: `select saveObject('report', NULL, '" + ...`. It's hard enough to read as is.

scripts/xml/add-manufacturing-to-distribution.xml
```diff
@@ -93,6 +93,18 @@
...
+                name = "Checking for duplicate Screens" >
+    <query>SELECT NOT EXISTS(SELECT 1
+                         FROM uiform a
+                         WHERE EXISTS(SELECT 1
+                                      FROM uiform b
+                                      WHERE a.uiform_name=b.uiform_name
+                                      AND a.uiform_order=b.uiform_order
+                                      AND a.uiform_id!=b.uiform_id));</query>
+    <message>There are multiple Screens (uiform table) with the same name and grade. Please see your system administrator or contact xTuple.</message>
+  </prerequisite>
```
gilmoskowitz: Ask the question more directly:
```sql
select not exists(select count(*), uiform_name, uiform_order
                    from uiform
                   group by uiform_name, uiform_order
                   having count(*) > 1);
```

foundation-database/public/functions/saveobject.sql
ChristopherCotnoir: It's selecting the lowest grade that doesn't exist in any table BUT the extension's, so FROM ONLY will not work.

foundation-database/public/functions/saveobject.sql
ChristopherCotnoir: No, because you can't have two reports/metasql/uiforms with the same grade even if they are in different packages. Scripts don't have this issue, so they can just update/insert at whatever order they are given in the package.xml/manifest.js. Here, we have to ensure that if an insert occurs, it occurs at a grade that does not already exist elsewhere. And we only want to update something if it exists at the grade it would be inserted at, otherwise we could overwrite custom metasql/reports/uiforms. And if we only updated if it existed at the grade passed, if the grade passed was already used, it would insert again every time the package is applied. This method is messy, but prevents all that. It picks the smallest possible grade it can insert at, subject to any explicit grades given to ensure they are followed, updates if it exists in that package at that grade to ensure idempotency, and inserts if it does not exist in that package at that grade, even if it exists in that package at another grade, because in those cases it is custom.

PR #2950, opened 2017-01-20
Issue #29239: Resolve display of PO items and non-inventory items in …

andersdd:
…PO screen


rkannadevara:

PR #2954, opened 2017-01-26
Issue #29293: Resolve inability to void Voucher. 

andersdd:
Caused by missing manifest entry for table changes in xtuple/xtuple#2923


lokeshsh:
Performed UAT: Issue is fixed

PR #2956, opened 2017-01-31
issue #29058 - reduce overhead finding next check number

gilmoskowitz:


sindhupriyaav:
Performed UAT: Following are the observations.

Tested in xTuple 4.10.0 masterref database upgraded from xTuple 4.9.5. Observed that application is hanged for few minutes while printing payment with large gap between next check number(say '3') and max check number(say '999999999'). However no error dialog observed and application is not terminated.

Tested in xTuple 4.11.0Beta pull request database. Observed that no hanging of the application is observed while printing the payment.

PR #2960, opened 2017-02-15
Fixes #29497. Get the ORM View primary key column, not base table.

bendiy:


sindhupriyaav:
Performed UAT: Observed that  'var pkeyColumn = XT.Orm.primaryKey(this.orm, true);' changed to 'var pkeyColumn = XT.Orm.primaryKey(this.orm);'  in 'data.sql' file.

PR #2962, opened 2017-02-16
Issue #29391: Resolve Div/0 error in Slow Moving Stk report

andersdd:
There's no point displaying slow moving items with no Qoh so this prevents the Div/0 potential at the same time


sindhupriyaav:
Performed UAT: Observed that items with 'QOH' as 0 are not displayed in 'Slow Moving Inventory' screen. Issue is fixed.

PR #2964, opened 2017-02-23
Fixes #29591 The `usr` view is slow on databases with lots of `usrpref` records

bendiy:
@mdathe tested this on a database with 400,000+ `usrpref` records and saw an order of magnitude performance increase.

### DIFF
--- a/foundation-database/public/views/usr.sql
+++ b/foundation-database/public/views/usr.sql
@@ -57,6 +57,8 @@ SELECT xt.create_view('public.usr', $BODY$
                   'propername',
                   'window'
                 ]) AS name
+               ORDER BY
+                name
             ) AS pref_names
             LEFT JOIN (
               -- Join the user's pref settings with the dummy table.
@@ -75,6 +77,9 @@ SELECT xt.create_view('public.usr', $BODY$
                    'propername',
                    'window'
                  )
+               ORDER BY
+                username,
+                name
             ) AS userprefs USING (name)
            -- Make sure they are alpha ordered so the array access operators above know which is which.
            ORDER BY
--- a/foundation-database/public/views/usr.sql
+++ b/foundation-database/public/views/usr.sql
@@ -1,20 +1,87 @@
+SELECT xt.create_view('public.usr', $BODY$
+-- Copyright (c) 1999-2017 by OpenMFG LLC, d/b/a xTuple.
+-- See www.xtuple.com/EULA for the full text of the software license.
 
---SELECT dropIfExists('VIEW', 'usr');
-CREATE OR REPLACE VIEW usr AS
-  SELECT usesysid::integer AS usr_id,
-         usename::text AS usr_username,
-         COALESCE((SELECT usrpref_value FROM usrpref WHERE usrpref_username=usename AND usrpref_name='propername'), '') AS usr_propername,
-         null::text AS usr_passwd,
-         COALESCE((SELECT CAST(usrpref_value AS INTEGER) FROM usrpref WHERE usrpref_username=usename AND usrpref_name='locale_id'),
-           COALESCE((SELECT locale_id FROM locale WHERE lower(locale_code) = 'default' LIMIT 1), (SELECT locale_id FROM locale ORDER BY locale_id LIMIT 1))) AS usr_locale_id,
-         COALESCE((SELECT usrpref_value FROM usrpref WHERE usrpref_username=usename AND usrpref_name='initials'), '') AS usr_initials,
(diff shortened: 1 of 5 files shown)


foundation-database/public/views/usr.sql
```diff
@@ -1,20 +1,87 @@
...
+  )
+  SELECT
+    usesysid::integer AS usr_id,
+    usename::text AS usr_username,
+    COALESCE(value[6], '') AS usr_propername,
+    null::text AS usr_passwd,
+    COALESCE(value[5]::integer, (SELECT locale_id FROM default_locale)) AS usr_locale_id,
+    COALESCE(value[4], '') AS usr_initials,
+    COALESCE((value[2] = 't'), false) AS usr_agent,
+    COALESCE((value[1] = 't'), usercanlogin(usename::text), false) AS usr_active,
```
bendiy: @gilmoskowitz an argument can be made that we don't need the call to `usercanlogin()` at all. It does slow the view done some.

scripts/install_xtuple.sh
```diff
@@ -160,13 +174,15 @@ install_packages() {
...
-  elif [ ${PG_VERSION} = 9.4 ] ; then
-    PLV8PKG="postgresql-${PG_VERSION}-plv8=1:1.4.8.ds-1.pgdg14.04+1"
-  fi
+
+# we had problems with a newer plv8 in mar-apr 2016, but it's working in Feb-2017.
+#  if [ ${PG_VERSION} = 9.3 ] ; then
+#    PLV8PKG="postgresql-${PG_VERSION}-plv8=1.4.0.ds-2"
+#  elif [ ${PG_VERSION} = 9.4 ] ; then
+#    PLV8PKG="postgresql-${PG_VERSION}-plv8=1:1.4.8.ds-2.pgdg14.04+1"
+     # Now installs (note: ds-2): postgresql-9.4-plv8 (1:1.4.8.ds-2.pgdg14.04+1) ...
```
bendiy: @gilmoskowitz If you look at the commits on this PR, I had no issue with pg 9.5, but 9.4 and 9.3 raised a bug. That is fixed with the `CROSS JOIN` in the last commit. When debugging, I found our issue with plv8 in pg 9.4 was that we referenced the wrong package here:
`plv8=1:1.4.8.ds-1.pgdg14.04+1`
vs.
`plv8=1:1.4.8.ds-2.pgdg14.04+1`
Note the `ds-2`. That is what will be installed now by default from `http://apt.postgresql.org/pub/repos/apt/`, so I don't think we need this spacial handling for 9.3 and 9.4 anymore. CI should now work for 9.3, 9.4 and 9.5. I think we should keep this at 9.3 since that is our minimum to protect against code added with features that are not in 9.3.

foundation-database/public/views/usr.sql
```diff
@@ -1,20 +1,101 @@
...
+  )
+  SELECT
+    usr_id,
+    usr_username,
+    COALESCE(value[6], '') AS usr_propername,
+    usr_passwd,
+    COALESCE(value[5]::integer, (SELECT locale_id FROM default_locale)) AS usr_locale_id,
+    COALESCE(value[4], '') AS usr_initials,
+    COALESCE((value[2] = 't'), false) AS usr_agent,
+    COALESCE((value[1] = 't'), usercanlogin(usr_username::text), false) AS usr_active,
```
gilmoskowitz: Get rid of the `userCanLogin()` call. It's doing a lot of work and may cause this view to return the wrong value for `usr_active`.

foundation-database/public/views/usr.sql
```diff
@@ -1,20 +1,101 @@
...
+      2 AS sort
+      FROM (
+        SELECT
+          locale_id
+          FROM locale
+         ORDER BY locale_id
+      ) AS locale_sort
+     ORDER BY sort
+     LIMIT 1
+  )
```
gilmoskowitz: This CTE can be greatly simplified with no performance impact:
```SQL
  WITH default_locale AS (
    SELECT locale_id, NOT (lower(locale_code) = 'default') AS sort
      FROM locale
     ORDER BY sort, locale_id
     LIMIT 1
  )
```

bendiy:
@gilmoskowitz Fixed.

bendiy:
@jcarlin This test is brittle:
https://github.com/xtuple/xtuple/blob/4_11_x/test/specs/activity.js#L52-L65
It relies on the list returning in a certain order. Do you know if it's possible to get a specific user from the list, `postgres`, instead of using:
```
      // Select the first model from the list, call the reassignUser function
      actList.select(0);
```

PR #2966, opened 2017-02-27
Issue #29361: SQL Error when voiding invoices

andersdd:


sindhupriyaav:

PR #2969, opened 2017-03-02
issue #25224 Eliminate duplicated query calls

ChristopherCotnoir:


ChristopherCotnoir:
Related to xtuple/qt-client#1532.

foundation-database/public/trigger_functions/coitem.sql
```diff
@@ -540,32 +540,42 @@ BEGIN
...
-        SELECT createPR(CAST(_coheadnumber AS INTEGER), 'S', NEW.coitem_id) INTO _orderid;
-        IF (_orderid > 0) THEN
-          UPDATE coitem SET coitem_order_id=_orderid
-          WHERE (coitem_id=NEW.coitem_id);
-        ELSE
-          RAISE EXCEPTION 'CreatePR failed, result=%', _orderid;
-        END IF;
+      SELECT createPR(CAST(cohead_number AS INTEGER), 'S', NEW.coitem_id) INTO _orderid
+      FROM cohead
+      WHERE (cohead_id=NEW.coitem_cohead_id);
```
gilmoskowitz: No need to `select` here as we already have `_r.cohead_number`.

foundation-database/public/trigger_functions/coitem.sql
```diff
@@ -540,32 +540,42 @@ BEGIN
...
+        UPDATE coitem SET coitem_order_id=_orderid
+        WHERE (coitem_id=NEW.coitem_id);
+
+        INSERT INTO charass
+        (charass_target_type, charass_target_id,
+         charass_char_id, charass_value)
+         SELECT 'R', NEW.coitem_order_id, charass_char_id, charass_value
+         FROM charass
+         WHERE ((charass_target_type='SI')
+         AND  (charass_target_id=NEW.coitem_id));
```
gilmoskowitz: Shouldn't the `coitem` update and `charass` handling be in `createpr()`? Check other calls to `createpr()` to see if there's a reason not to move the logic.

foundation-database/public/trigger_functions/coitem.sql
```diff
@@ -540,32 +540,42 @@ BEGIN
...
       END IF;
     END IF;
 
     -- Create Purchase Order if flagged to do so
     IF ((NEW.coitem_order_type='P') AND (NEW.coitem_order_id=-1)) THEN
       SELECT itemsrc_id INTO _itemsrcid
-      FROM itemsite JOIN itemsrc ON (itemsrc_item_id=itemsite_item_id AND itemsrc_default)
+      FROM itemsite JOIN itemsrc ON (itemsrc_item_id=itemsite_item_id AND itemsrc_default AND itemsrc_active)
       WHERE (itemsite_id=NEW.coitem_itemsite_id)
-        AND (NOT itemsite_stocked);
```
gilmoskowitz: Check with @mdathe to see if replacing the `item_stocked` check is correct.

foundation-database/public/trigger_functions/coitem.sql
ChristopherCotnoir: createpr doesn't necessarily use a coitem as the parent, it also can use womatl or planord, which don't have characteristics. The logic is correct where it is.

ChristopherCotnoir:
I don't think splitting the trigger into two will work that well because of the nature of the shared logic, unless the auxiliary function is another trigger function (so have an upsert trigger, insert trigger, and update trigger) because a stored procedure would need an awful lot of parameters passed to it and wouldn't really be performing something coherent outside the context of the trigger.

foundation-database/public/trigger_functions/coitem.sql
ChristopherCotnoir: Confirmed with him that this is a wildly incorrect latent bug.

PR #2971, opened 2017-03-05
Issue #29408: Deleting Std cost did not update itemsite or history

andersdd:
This was a timing issue with calculating Std Cost.  Also affected the GL postings and Inventory History (not mentioned in original issue).

### DIFF
--- a/foundation-database/public/functions/updatestdcost.sql
+++ b/foundation-database/public/functions/updatestdcost.sql
@@ -3,7 +3,7 @@ CREATE OR REPLACE FUNCTION updateStdCost(pItemcostid INTEGER,
                                          pOldcost NUMERIC,
                                          pDocNumber TEXT,
                                          pNotes TEXT) RETURNS BOOLEAN AS $$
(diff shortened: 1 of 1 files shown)


lokeshsh:
Performed UAT: Please find the steps below:
Deleted the Standard cost for the item say 'Sheet1' from 'Maintain Item Costs' screen and observed that the value of 'itemsite_value' in the 'item site' table is not updated to '0'

gilmoskowitz:
@andersdd - the code looks correct but the UAT says otherwise. any ideas? this might be a good time to add unit tests for this function.

andersdd:
@gilmoskowitz @lokeshsh  UAT was correct.  The original change did correctly update std cost but not the Item Site.  I have corrected the function to correctly update both areas of cost.

andersdd:
@gilmoskowitz  I was just reviewing the current updateitemcost test and it looks wrong.  I don't know how it is passing as the wrong parameters are being passed into the function by the test.  Do you want me to go through and 'fix' it?

gilmoskowitz:
@andersdd - If you have time, fixing the test would be very helpful.

PR #2973, opened 2017-03-06
Issue #28773: Add Customer PO to Open Receivables

andersdd:
Added the same field logic to the formatted query but not to the report as I have left this to individual customers to decide if needed.


gilmoskowitz:
👍 when CI passes

PR #3009, opened 2017-03-28
issue #29788 - itemlocdist_source_id_idx - 15min to 1.5min

gilmoskowitz:
Adding an index on the source id improves performance significantly.
Adding a foreign key constraint on itemlocdist_itemlocdist_id_fkey
should also help but exposes a bug in inventory distribution
cleanup.
Simplifying the formatlocationname function speeds it up by
about 10% (not much but the code is much easier to read now).

### DIFF
--- a/foundation-database/public/functions/formatlocationname.sql
+++ b/foundation-database/public/functions/formatlocationname.sql
@@ -12,7 +12,7 @@ BEGIN
     FROM location
    WHERE location_id = pLocationid;
 
-  RETURN COALESCE(_name, 'NA');
+  RETURN COALESCE(_name, 'N/A');
 
 END;
 $$ LANGUAGE plpgsql;
--- a/foundation-database/public/functions/formatlocationname.sql
+++ b/foundation-database/public/functions/formatlocationname.sql
@@ -1,42 +1,19 @@
 
-CREATE OR REPLACE FUNCTION formatLocationName(INTEGER) RETURNS TEXT IMMUTABLE AS '
--- Copyright (c) 1999-2014 by OpenMFG LLC, d/b/a xTuple. 
+CREATE OR REPLACE FUNCTION formatLocationName(pLocationid INTEGER) RETURNS TEXT IMMUTABLE AS $$
+-- Copyright (c) 1999-2017 by OpenMFG LLC, d/b/a xTuple. 
 -- See www.xtuple.com/CPAL for the full text of the software license.
 DECLARE
-  pLocationid ALIAS FOR $1;
   _name TEXT;
-  _r RECORD;
 
 BEGIN
+  SELECT COALESCE(location_aisle, '') || COALESCE(location_rack, '')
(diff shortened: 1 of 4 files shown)


foundation-database/public/tables/itemlocdist.sql
```diff
@@ -1,5 +1,25 @@
...
+       xt.add_column('itemlocdist', 'itemlocdist_expiration',     'DATE',          NULL, 'public'),
+       xt.add_column('itemlocdist', 'itemlocdist_distlotserial',  'BOOLEAN',       NULL, 'public'),
+       xt.add_column('itemlocdist', 'itemlocdist_warranty',       'DATE',          NULL, 'public'),
+       xt.add_column('itemlocdist', 'itemlocdist_ls_id',          'INTEGER',       NULL, 'public'),
+       xt.add_column('itemlocdist', 'itemlocdist_order_type',     'TEXT',          NULL, 'public'),
+       xt.add_column('itemlocdist', 'itemlocdist_order_id',       'INTEGER',       NULL, 'public'),
+       xt.add_column('itemlocdist', 'itemlocdist_child_series',   'INTEGER',       NULL, 'public'),
+       xt.add_column('itemlocdist', 'itemlocdist_transtype',      'TEXT',          NULL, 'public');
+
+--SELECT xt.add_constraint('itemlocdist', 'itemlocdist_itemlocdist_id_fkey', 'FOREIGN KEY (itemlocdist_itemlocdist_id) REFERENCES itemlocdist(itemlocdist_id)', 'public');
```
jcarlin: This is resolved in https://github.com/xtuple/xtuple/pull/3006.

foundation-database/public/indexes/itemlocdist.sql
```diff
@@ -0,0 +1 @@
+select xt.add_index('itemlocdist', 'itemlocdist_source_id', 'itemlocdist_source_id_idx', 'btree', 'public');
```
jcarlin: `itemlocdist_source_id` can be an `itemlocdist_id` or a `location_id`, making it possible that there could be duplicate values in `itemlocdist_source_id`. In fact, even with the case of `itemloc_source_id` being a `itemlocdist_id` value, there could be duplicate values for this column. This index would throw an error in either case I believe.

gilmoskowitz: There's no `UNIQUE` constraint on the index so both cases should be fine.

xtuple-ci-user:
Build finished.

foundation-database/public/functions/formatlocationname.sql
```diff
@@ -1,42 +1,19 @@
...
-
-    IF (_r.location_name IS NOT NULL) THEN
-      _name := (_name || _r.location_name);
-    END IF;
-
-    RETURN _name;
-  ELSE
-    RETURN ''N/A'';
-  END IF;
+  RETURN COALESCE(_name, 'NA');
```
ChristopherCotnoir: Should NA still be N/A?

PR #3014, opened 2017-04-20
issue #29885 Fix regression from issue #28597 and patch database

ChristopherCotnoir:


lokeshsh:

PR #3019, opened 2017-04-27
issue #29907 Fix error in conditions for fetchnextchecknumber

ChristopherCotnoir:


harithamekala:
Performed UAT: Observed that next payment number is not displayed with old sequence. But observed the following behavior:

1.Open 'Bank Account' say 'EBANK' and observed that next payment number as '4180'.

2. Create & post miscellaneous payment for the above bank and observed that payment number is '4180'.

3. Now change the system date to future date say '01-Jan-2018' and login to the application.

4. Change the bank account 'EBANK' next payment number to '2180' which is lower than the old sequence '4180'.

5. Now create a miscellaneous payment for above bank 'EBANK'.

6. Observed that payment number is displayed as '2180' in 'Print Payment' window.

7. Now click on 'Already Printed/Paid' and observed that payment number is displayed as '3050' instead of '2180' in 'View Payment Run' screen.### Please refer the below screenshot #####
screenshot 12

PR #3022, opened 2017-05-07
Issue #29827: Correct demo item characteristic settings

andersdd:
Allow correct settings for item characteristics in Demo database data.


andersdd:
Some build timing issues require resolution

PR #3036, opened 2017-05-19
issue #30126 Show Purchase Orders without a type

ChristopherCotnoir:


ChristopherCotnoir:
Associated with xtuple/qt-client#1600.

PR #3040, opened 2017-05-24
issue #30171 Don't assume no record exists on insert

ChristopherCotnoir:


lib/orm/source/xt/trigger_functions/record_did_change.sql
```diff
@@ -37,20 +36,14 @@ return (function () {
...
-     }
-   }
-
+   /* find a version record, if found increment */
+   sql = 'select ver_id from xt.ver where ver_table_oid = $1 and ver_record_id = $2;';
+   qry = plv8.execute(sql, [oid, NEW[pkey]]);
+   if (qry.length) {
+     sql = 'update xt.ver set ver_etag = $1::uuid where ver_id = $2;';
+     plv8.execute(sql, [ XT.generateUUID(), qry[0].ver_id - 0]);
+   } else {
```
gilmoskowitz: Simplify even further by removing the `select`:
```
qry = UPDATE ... RETURNING ver_id;
if (! qry.length)
    INSERT ...;
```

PR #3041, opened 2017-05-24
issue #30161 Added unique constraint to potype table

ChristopherCotnoir:


ChristopherCotnoir:
Requires xtuple/qt-client#1601.

PR #3045, opened 2017-05-26
Fixes #30183 Remove html5 module and fix it in a fork of zombie.

bendiy:
Our `npm install`s are now broken.

Being a an old node version `0.10.40` has now bit us.

You may see:
```npm ERR! 404 Not Found
npm ERR! 404 
npm ERR! 404 'types/node' is not in the npm registry.
npm ERR! 404 You should bug the author to publish it
npm ERR! 404 It was specified as a dependency of 'parse5'
npm ERR! 404 
npm ERR! 404 Note that you can also install from a
npm ERR! 404 tarball, folder, or http url, or git url.

npm ERR! System Linux 3.13.0-44-generic
npm ERR! command "node" "/usr/local/bin/npm" "install"
npm ERR! cwd /opt/xtuple/4.11.0-beta/nn410/xtuple
npm ERR! node -v v0.10.40
npm ERR! npm -v 1.4.28
npm ERR! code E404
npm ERR! not ok code 0
```

or

```npm WARN engine co@4.6.0: wanted: {"iojs":">= 1.0.0","node":">= 0.12.0"} (current: {"node":"0.10.40","npm":"1.4.28"})
npm ERR! cb() never called!
npm ERR! not ok code 0
```

That will happen if you are on `npm -v` version `1.4.28`.
The fix is to upgrade to npm version 2. version 3 might work, didn't try. version 4 is too new and does not run on node `0.10.40` (edited)
You can do this by running:

`npm install -g npm@2` as `root`

or

`sudo npm install -g npm@2`

Then `npm install` will work.

Root cause is:
We have the `html5` modules in our dev dependencies:
https://github.com/xtuple/xtuple/blob/4_11_x/package.json#L46
The version we use requires the `jsdom` package:
https://github.com/aredridel/html5/blob/v0.3.13/package.json#L42
But they used `>= 0.6.0`, which means any version greater than or equal to `0.6.0`.
The `jsdom` package just released version `11.0.0`, which includes the latest version of the `parse5` modules:
https://github.com/tmpvar/jsdom/blame/master/package.json#L30
The new `parse5` module `v3.0.2` includes a `"@types/node": "^6.0.46"` dependency, which is new syntax that `npm -v 1.4.28` does not understand:
https://github.com/inikulin/parse5/blob/v3.0.2/package.json#L61

`npm -v 2.x` does understand that new syntax.
 

Something else that might work is setting the `export NODE_ENV=production` so `npm` will skip `devDependencies`
The newer versions of `html5` use `"jsdom": "^0.11.0"` which will keep it on the right version. https://github.com/aredridel/html5/blob/v1.0.5/package.json#L41

Unfortunately, this is going to break any xtuple installs. Including the old ones. 4.10, 4.9 and before.


gilmoskowitz:
+1 when jenkins passes

bendiy:
It gets worst.  Our tests use `zombie`, which is what actually needed this `html5` module. We did not use the `html5` module directly and there's no reason it should be in _our_ `package.json`. We should be letting `zombie` install it.

But.... `zombie` version `1.4.1` has the same issue with `html5`. Which is removed in `zombie` version `2.0.0`, but... that version is completely different. So I can remove `html5` from _our_ `package.json` without an issue. But `zombie` version `1.4.1` will install it and we'll see the CI failure in my PR.

I can update `zombie`'s nested `html5` module like I did to ours so it is version `1.0.x` and then the tests run fine. But there's no way to do that other than manually.
```rm -rf node_modules/zombie/node_modules/html5
cd node_modules/zombie
npm install html5@1.0.x
cd ../..```

**Comment by gilmoskowitz** [comment]
Is this sufficient? What about [the shrinkwrap file](https://github.com/xtuple/xtuple/blob/4_11_x/npm-shrinkwrap.json)?

**Comment by bendiy** [comment]
Shrinkwrap doesn't apply to `devDependencies` in `package.json`.

**Comment by bendiy** [comment]
```
Starting with npm v4.0.1, devDependencies are included when you run npm shrinkwrap and follow the usual rules as to when they're installed. As of npm v3.10.8, if you run npm install --only=production or npm install --production with a shrinkwrap including your development dependencies they won't be installed. Similarly, if the environment variable NODE_ENV is production then they won't be installed. If you need compatibility with versions of npm prior to v3.10.8 or otherwise don't want them in your shrinkwrap you can exclude development dependencies with: npm shrinkwrap --only=prod or npm shrinkwrap --production.
```
We're still using the old npm version.

PR #3051, opened 2017-06-07
issue #30186 - disable triggers during schema changes

gilmoskowitz:
- removed some redundancy on the way to idempotency
- removed a obsolete table
- removed an uncalled trigger function
- added creation of some indexes somehow missed in previous changes
- tweaked some outdated table and column comments

Manually tested with `pg_dump --schema-only` of a web-enabled database created with and without these changes.


gilmoskowitz:
Note to reviewers: This is probably best reviewed locally as the diff for 440_schema.sql is too big to show on github. Use `git diff --diff-algorithm=minimal` instead of the default algorithm - the diff is about an order of magnitude smaller that way:
```
$ for i in patience minimal histogram myers ; do
>   echo $i: $(git diff --diff-algorithm=$i e319b8404e 42ad1717ea | wc)
> done
patience: 26952 122584 1015780
minimal: 26950 122580 1015776
histogram: 205914 954292 8187038
myers: 205914 954292 8187038
```

PR #3057, opened 2017-06-16
Add indexes to core tables used by xTupleCommerce.

bendiy:
This PR adds indexes on columns used for `JOIN`s, `WHERE`s and in searches by xTupleCommerce.


foundation-database/public/indexes/cohead.sql
```diff
@@ -0,0 +1,20 @@
+-- Using `text_pattern_ops` indexes on text columns to improve search speed.
+-- Change old index to `text_pattern_ops`.
```
gilmoskowitz: Do we know that regex searching is more frequently used in practice? Here and elsewhere (e.g. `cntct`).

gilmoskowitz: Some experimentation in a different context showed that text_pattern_ops did not in practice cause a regex search to use the index. We need to verify that these indexes will be used before creating them.

bendiy: @gilmoskowitz @davidbeauchamp 
Today, I needed to query for any Contacts with a matching email address. This lead me to test some of the different indexes on `cntct` proposed here. Most of our MetaSQL is using `~*` to support regex pattern matching.

With a query like:
```sql
SELECT
  cntct_id
FROM cntct
WHERE true
  AND cntct_email ~* 'foo@example.com';
```
I get a `Seq scan`, even with our existing index:
```sql
SELECT xt.add_index('cntct', 'cntct_email', 'cntct_email_idx', 'btree', 'public');
```
It takes 3 seconds on a large database.

Switching to a `text_pattern_ops` did not help. But, I found this:
https://dba.stackexchange.com/a/10696
Relevant part:
> Such an index is only good for left-anchored patterns (matching from the start of the string). ... and simple regexp patterns with ~

If I add the `^` character, the current index is still not hit, but query time drops to 1.1 seconds.

If I use a `text_pattern_ops` index, the index is still not hit, but query time is the same 1.1 seconds.

If I switch to a case sensitive `~` match and include `^`, I hit the index. 
```sql
SELECT
  cntct_id
FROM cntct
WHERE true
  AND cntct_email ~ '^foo@example.com';
```
43 ms

Takeaway, `text_pattern_ops` indexes are only going to be useful if we use `~`, not `~*` and if we left-anchor the pattern. However, if we do do that, they are very fast.

I still have some work to do on the PR when I can get back to it.

bendiy: Unfortunately, a functional index for `LOWER(cntct_email)` and/or using `LOWER()` in the WHERE clause doesn't help.
```sql
SELECT
  cntct_id
FROM cntct
WHERE true
  AND LOWER(cntct_email) ~ '^foo@example.com';
```
Query doesn't use the `text_pattern_ops` index and goes back to 1+ seconds.

andersdd: @bendiy what about broader searches such as `show me all contacts with this company email` for the equivalent of searching all contacts at `*.xtuple.com`? You can't left anchor then.  One thing I have had quite good results with is a gin index where we convert the text to ts_vector.  I have also combined several text fields into the one index to search from (firstname, surname, phone, email).  Have that working on a contacts database with several hundred thousand contacts in.

bendiy: @andersdd 
That is better:
```sql
CREATE EXTENSION IF NOT EXISTS pg_trgm;
SELECT xt.add_index('cntct', 'cntct_email gin_trgm_ops', 'cntct_email_idx', 'gin', 'public');
```
```sql
SELECT
  cntct_id
FROM cntct
WHERE true
  AND cntct_email ~* 'foo@example.com';
```
463 ms vs. the original 3 seconds.

gilmoskowitz: An order of magnitude is much better and worth pursuing. As long as `pg_trgm` is supplied such that `create extension` line is all we need to do, I'm good with it. If any additional steps are needed, such as building and installing the extension, we need to consider carefully - we'll be imposing yet another barrier to entry.

gilmoskowitz:
This set of changes is too big and not sufficiently tested to merge. Smaller change sets, tested incrementally, with demonstrated performance improvements to in-the-field performance problems will be more palatable.

PR #3059, opened 2017-06-16
Fixes #30252: listprice() should work for exclusive items too.

bendiy:
With this change `listprice()` will return a "List Price" from the `item.item_listprice` column for `item_exclusive === true` Items.

`salesOrderItem.cpp` expects a `-9999.0` value, but it gets it from `itemipsprice()`, not `listprice()`:
https://github.com/xtuple/xtuple/blob/4_11_x/foundation-database/public/functions/itemipsprice.sql#L206


lokeshsh:
Performed UAT: Issue is fixed

PR #3062, opened 2017-06-20
Add subnumber to invoice

ChristopherCotnoir:


ChristopherCotnoir:
Associated with xtuple/enhanced-pricing#30

jcarlin:
Add `item_type`? Sales Order and Quote item list metasql both have it. Also, is `soitemnumber` sufficient for a formatted subnumber column (it uses `coitem_subnumber`) or should there be a new formatted sub number column using `invcitem_subnumber` as well? i.e. https://github.com/xtuple/xtuple/pull/3055/files

ChristopherCotnoir:
There is no point in adding columns to the metasql if the client isn't looking for those columns. Remember that invoices are different from Sales Orders and Quotes, which are much more similar to each other, it is expected that the columns will not be the same. You can ask Ben, but I'm pretty sure using coitem's formatting is fine.

jcarlin:
OK. We _will_ need new formatting when we allow discount codes on invoices without sales orders.

PR #3064, opened 2017-06-22
Carry coitem subnumbers over to invcitem

ChristopherCotnoir:


foundation-database/public/functions/createinvoice.sql
```diff
@@ -150,14 +151,13 @@ BEGIN
...
     FROM invcitem, coitem, cobmisc
-    WHERE ((invcitem_linenumber=_lastlinenumber)
+    WHERE ((invcitem_linenumber=coitem_linenumber)
+      AND  (invcitem_subnumber=coitem_subnumber)
       AND  (coitem_id=cobill_coitem_id)
       AND  (cobmisc_id=cobill_cobmisc_id)
       AND  (cobill_cobmisc_id=pCobmiscid)
       AND  (invcitem_invchead_id=_invcheadid));
     
-    _lastlinenumber := _lastlinenumber + 1;
```
jcarlin: Changing this will result in skipped line numbers in the case of deleted sales order line items.

ChristopherCotnoir: If we're going to be adding subnumbers to invoices, it seems to me there's no longer any reason to make invoice line numbers different from their originating sales order. I think it makes more sense for the invoice to match the sales order, even if numbers may be skipped.

jcarlin: Let's ask @gilmoskowitz. I know that some customers would be very confused if the line number sequence skipped on their invoice.

gilmoskowitz: I have a slight preference for sequential line numbers instead of matching the sales order. Rationale:
- skipping might be confusing
- skipping isn't necessary
- the invoice lines are created in coitem_linenumber order so side-by-side comparison is simple

foundation-database/public/functions/createinvoice.sql
```diff
@@ -93,6 +96,15 @@ BEGIN
...
 
+    IF (_lastcoitemline!=_r.coitem_linenumber) THEN
+      _lastlinenumber := _lastlinenumber + 1;
+      _lastsubnumber := 0;
+    ELSE
+      _lastsubnumber := _lastsubnumber + 1;
+    END IF;
+
+    _lastcoitemline = _r.coitem_linenumber;
+
```
jcarlin: Nice.

PR #3066, opened 2017-06-23
issue #30405 Don't ignore invoice items with negative price when posting (for disc…

ChristopherCotnoir:
…ounts)

### DIFF
--- a/foundation-database/public/functions/postinvoice.sql
+++ b/foundation-database/public/functions/postinvoice.sql
@@ -151,7 +151,7 @@ BEGIN
 --  Cache the amount due for this line
     _amount := _r.extprice;
 
-    IF (_amount > 0) THEN
+    IF (_amount != 0) THEN
 --  Credit the Sales Account for the invcitem item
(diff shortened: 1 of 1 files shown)


foundation-database/public/functions/postinvoice.sql
```diff
@@ -151,7 +151,7 @@ BEGIN
 --  Cache the amount due for this line
     _amount := _r.extprice;
 
-    IF (_amount > 0) THEN
+    IF (_amount != 0) THEN
```
gilmoskowitz: Check with @mccjim - Do we want to "credit" the discount to the same account or allow for a distinct discount account in the G/L?

mccjim: So my recommendation would be if the calculated line item amount is less than zero then use the sales assignment for "Sales Credit Account" but honoring the "Alternate Revenue Account" if provided.

The Sales assignments give you enough options to code the item which is being credited.

PR #3086, opened 2017-07-13
DO NOT MERGE - issue #30536 Properly schema qualify function

ChristopherCotnoir:


foundation-database/public/functions/getbooitemseqid.sql
```diff
@@ -16,7 +16,7 @@ BEGIN
     RETURN -1;
   ELSE
     SELECT booitem_seq_id INTO _returnVal
-    FROM booitem(getItemId(pItemNumber))
+    FROM xtmfg.booitem(getItemId(pItemNumber))
```
bendiy: Why isn't this picked up through the search path?

rkannadevara:
Performed UAT: Issue is fixed. Observed that no error is displayed and  ‘20’ value is displayed when ‘SELECT getbooitemseqid('BRCPFN1',’10’);’ query is executed. Please refer to the below attached screenshot.

30536issue_fixed

foundation-database/public/functions/getbooitemseqid.sql
ChristopherCotnoir: Because this function is never run from the client, it is only called on insert/update to api.bomitem. If run through the client, I suspect it would work, but running from pgadmin or psql will not set the search path. This is why functions should always be schema qualified.

bendiy: OK. I'm going to hold off on this and defer to @gilmoskowitz.

If the use case is a non-xtuple client doing an import, that client needs to call our init functions that will set the search path. 
```
SELECT xt.js_init();
-- OR
SELECT login(true);
```
We intentionally do not schema qualify function so that extensions can override them. Using the correct search path setting we make through those init functions.

gilmoskowitz: What Ben said:
> We intentionally do not schema qualify functions so that extensions can override them.

That's the goal even if it's poorly implemented in practice. We might want to modify  OpenRPT and CSVImp to call `login(true)` IFF it's defined.

ChristopherCotnoir:
Fixing with xtuple/openrpt#78 instead. Closing this.

PR #3087, opened 2017-07-14
Issue #30537: Remove recur constraint as it was preventing creation

andersdd:
of any recurring document

### DIFF
--- a/foundation-database/public/tables/recur.sql
+++ b/foundation-database/public/tables/recur.sql
@@ -13,11 +13,10 @@ select xt.add_column('recur','recur_max', 'INTEGER', '', 'public');
 select xt.add_column('recur','recur_data', 'TEXT', '', 'public');
 select xt.add_column('recur','recur_style', 'TEXT', 'DEFAULT ''KeepNone'' NOT NULL', 'public');
 
-SELECT
-  xt.add_constraint('recur', 'recur_recur_period_check',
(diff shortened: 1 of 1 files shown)


gilmoskowitz:
This is an important constraint that's been functioning properly for years. There's got to be something else going on that's causing the f-key violation.

bendiy:
@gilmoskowitz 
It didn't exist 10 months ago:
https://github.com/andersdd/xtuple/blame/ba606674e52fbbb1d696ff3836bddcfd09232d73/foundation-database/public/tables/recur.sql

`COMMENT ON COLUMN recur.recur_parent_id IS 'The internal ID of the event/object that recurs.';`
`COMMENT ON COLUMN recur.recur_parent_type IS 'The table in which the parent event or object is stored.';`

Dave and my understanding was that with a `recur_parent_type ` of `I`, `recur_parent_id` would reference an `invchead_id`, not a `recur_id`. Therefore, it's not an fkey to `recur_id`.

andersdd:
@bendiy @gilmoskowitz Ben is correct - the ID is referencing the source document (Invoice/Voucher id) which is why this constraint was failing.  The constraint, as it stood, was wrong and most definitely the cause of this issue.

gilmoskowitz:
Doh! I was thinking of a different column. Mea culpa.

PR #3090, opened 2017-07-19
Issue #29317 Handle poor parameter filter value in GL Transactions report

andersdd:


harithamekala:
Performed UAT: Issue is fixed. No DB Log Error exists. Please refer to the attached screenshot ‘GlTrans.png’
gltrans

PR #3092, opened 2017-07-21
First pass at translation refactor

ChristopherCotnoir:


ChristopherCotnoir:
Associated with xtuple/updater#51 and xtuple/qt-client#1672

ChristopherCotnoir:
Test is failing because there is already an xt.dict table. Not sure what the best course of action is; naming the tables dict/pkgdict seems the most sensible long term solution, because it matches the name of the standard directory structures, just like with metasql/script/uiform/report, and we probably don't want that old table anymore anyway. Should we add another exception to gotchas.js like with the potype table, or rename this, and if renaming, what should it be named? We could name it back to qm, but if we are going to treat this as a true package object rather than just a collective place to store all qm like I was originally doing, it seems awkward.

scripts/release_build.sh
```diff
@@ -203,6 +208,28 @@ EDITIONS="postbooks manufacturing distribution"
...
+lupdate -no-obsolete xtuple.pro
+lrelease xtuple.pro
+mv share/dict/*.qm ../xtuple/scripts/output/dict/postbooks
+
+cd ../private-extensions
+for PACKAGE in $EDITIONS $PACKAGES ; do
+  if [ "$PACKAGE" != postbooks -a "$PACKAGE" != inventory ] ; then
+    lupdate -no-obsolete source/$PACKAGE/foundation-database/*/tables/dict/*_ts.pro
+    lrelease source/$PACKAGE/foundation-database/*/tables/dict/*_ts.pro
+    cp source/$PACKAGE/foundation-database/*/tables/dict/*.qm ../xtuple/scripts/output/dict/$PACKAGE
```
gilmoskowitz: `mv` would be preferable.

scripts/release_build.sh
```diff
@@ -271,6 +285,10 @@ for EDITION in $EDITIONS enterprise ; do
             cp scripts/output/$SUBPACKAGE-$SUBMODE.sql scripts/output/$FULLNAME
           fi
         done
+
+        if [ "$SUBPACKAGE" != inventory ] ; then
+          cp scripts/output/dict/$SUBPACKAGE/*.qm scripts/output/$FULLNAME
```
gilmoskowitz: `mv` would be preferable.

It's unclear why the inventory extension is skipped here. If the reason is that it has no `dict` subdirectory, make the script more robust by looking for `dict` instead of having a static name check. If there's some other reason, add an explanatory comment.

scripts/release_build.sh
```diff
@@ -281,6 +299,37 @@ done
...
 
+# build updater so we can use it
+cd ../qt-client
+git submodule update --init --recursive
+cd openrpt
+qmake
+make
+cd ..
+qmake
+make
```
gilmoskowitz: There's no need to build all of `qt-client`, just `common`.

scripts/release_build.sh
```diff
@@ -203,6 +208,28 @@ EDITIONS="postbooks manufacturing distribution"
...
 PACKAGES="inventory commercialcore"
 
+# translations
+mkdir -p $XTUPLEDIR/scripts/output/dict/postbooks
+mkdir -p $XTUPLEDIR/scripts/output/dict/commercialcore
+mkdir -p $XTUPLEDIR/scripts/output/dict/manufacturing
+mkdir -p $XTUPLEDIR/scripts/output/dict/distribution
+
+cd ../qt-client
+lupdate -no-obsolete xtuple.pro
```
gilmoskowitz: May need to `make clean` before this and move openrpt and csvimp out of the way. Otherwise the translation files may get stuffed with info we don't need.

scripts/xml/distribution-install.xml
```diff
@@ -124,5 +124,89 @@
...
+  <loadqm file="xTuple.pl.qm"/>
+  <loadqm file="xTuple.pt_br.qm"/>
+  <loadqm file="xTuple.pt.qm"/>
+  <loadqm file="xTuple.ru.qm"/>
+  <loadqm file="xTuple.sk.qm"/>
+  <loadqm file="xTuple.tr.qm"/>
+  <loadqm file="xTuple.uk.qm"/>
+  <loadqm file="xTuple.zh_hk.qm"/>
+  <loadqm file="xTuple.zh.qm"/>
+  <loadqm file="xTuple.zh_tw.qm"/>
```
gilmoskowitz: We should consider using `xmllint --xinclude` to build these .xml files from shared component files, such as `<loadqm\>` lists and prerequisite checks.

scripts/release_build.sh
ChristopherCotnoir: Not anymore. This was an issue when we generated a pro file with qmake and used that to generate translations. But there was no need for that roundabout method, so I changed the xtuple.pro file to include the translations so that we can just use the same pro file we use for building. See the qt-client PR.

scripts/release_build.sh
ChristopherCotnoir: mv does not work here, because these files are copied into multiple places throughout the loop, just like with the sql files. The directory is deleted at the end of the script anyway.

inventory is skipped because it has no dict, yes. I can change it to a check for dict throughout, though.

PR #3095, opened 2017-07-24
Issue #30391: Duplicate SO Items lines when similar alias codes are in system

andersdd:
This PR resolves stated issue, but there are a few other problems with aliases found during testing, possibly not documented in issues.


harithamekala:
Performed UAT: Issue is fixed, Observed only one line item is created when same alias is given for two different customers. Please refer to the attached screenshot Itemalias_sales.png’

itemalias_sales

PR #3108, opened 2017-07-31
Fixes: #30663 Force all pg connections to get a valid search path by calling destroy() on any pg connection pool client connections that generated an error in the database.

bendiy:
I suspect this may be the root cause of some of the nondeterministic errors we have been seeing in tests.

### DIFF
--- a/node-datasource/lib/ext/datasource.js
+++ b/node-datasource/lib/ext/datasource.js
@@ -16,7 +16,30 @@ Backbone:true, _:true, X:true, __dirname:true, exports:true, module: true */
     EventEmitter = require('events').EventEmitter,
     DataSource = {
       requestNum: 0,
-      callbacks: { },
+      callbacks: {},
+
+      creds : {},
+
+      /**
+       * The package, `node-postgres`, we use, exposed as `X.pg` maintains a
+       * connection pool to the database server, defaulting to 10 connections.
+       *
+       * We make use of plv8's `SET plv8.start_proc = 'xt.js_init' feature.
+       * This is only called the FIRST time a plv8 function is called for a
+       * connection. Because of this, if a connection, when added to the pool
+       * happens to cause an error in the database during its first query, all
+       * of the steps we do for a persistent client connection in
+       * `xt.js_init();` are rolled back. The main thing we setup is the
+       * `search_path`. If an error happens on the first query for this
+       * connection, the `search_path` will not be set the next time the pool's
+       * conneciton is used for a query.
+       *
+       * Because of this, on any emitted errors below in `connected()`, we
+       * destroy that client from the pool.
+       */
+      destroyClientFromPool: function (client) {
+        X.pg.pools.all[JSON.stringify(this.creds)].destroy(client);
+      },
 
       getAdminCredentials: function (organization) {
         return {
@@ -89,7 +112,7 @@ Backbone:true, _:true, X:true, __dirname:true, exports:true, module: true */
         @param {Function} callback
       */
       query: function (query, options, callback) {
-        var creds = {
+        this.creds = {
           "user": options.user,
           "port": options.port,
           "host": options.host || options.hostname,
@@ -103,7 +126,7 @@ Backbone:true, _:true, X:true, __dirname:true, exports:true, module: true */
           this.callbacks[this.requestNum] = callback;
           // Single worker version.
           options.debugDatabase = X.options.datasource.debugDatabase;
-          this.worker.send({id: this.requestNum, query: query, options: options, creds: creds, poolSize: this.poolSize});
+          this.worker.send({id: this.requestNum, query: query, options: options, creds: this.creds, poolSize: this.poolSize});
 
           // NOTE: Round robin benchmarks are slower then the above single pgworker code.
           // Round robin workers version. This might be useful in the future.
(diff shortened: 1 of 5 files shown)


bendiy:
jenkins - retest this please

bendiy:
I don't understand why this just started failing.
```
◦ needs a bankrecimport to process: [2K[0G    ✓ needs a bankrecimport to process 
    ◦ should succeed with various bankrecimports: [31m<<ERROR 2017-07-31T22:38:59.862Z>> [39m[31mDatabase Error! [39m[31mbankrec already posted Please fix this!!![39m
[31m<<ERROR 2017-07-31T22:38:59.862Z>> [39m[31mDatabase Error! Last query was: [39m[31mselect importBankrecCleared($1) as result;[39m
[31m<<ERROR 2017-07-31T22:38:59.862Z>> [39m[31mDatabase Error! DB name = [39m[31mjenkins3973[39m
[2K[0G    1) should succeed with various bankrecimports
    - needs a posted bankrec
    - should fail with posted bankrec
    ◦ needs to restore those metrics: [31m<<ERROR 2017-07-31T22:39:19.910Z>> [39m[31mDatabase Error! [39m[31mbankrec already posted Please fix this!!![39m
[31m<<ERROR 2017-07-31T22:39:19.910Z>> [39m[31mDatabase Error! Last query was: [39m[31mselect importBankrecCleared($1) as result;[39m
```

Why would we pass `bankrecimport.bankrecimport_id` to `importBankrecCleared(bankrecimport_id)`:
https://github.com/xtuple/xtuple/blob/4_11_x/test/database/functions/importbankreccleared.js#L184

and query `bankrec` expecting `WHERE (bankrec_id=bankrecimport_id)` to match?:
```
FROM bankrec JOIN bankaccnt ON (bankaccnt_id=bankrec_bankaccnt_id)
  WHERE (bankrec_id=_bankrecid);
```
https://github.com/xtuple/xtuple/blob/2fcb49fee3b83e0adbb371120ecb99a25d3cadc8/foundation-database/public/functions/importbankreccleared.sql#L27

`bankrec_id` and `bankrecimport_id` are completely different primary keys on different table and don't share a sequence.

test/database/dblib.js
```diff
@@ -134,11 +132,6 @@
           });
         };
     rgp();
-    rup();
-    dxtu();
-    du();
-
-    if (_.isFunction(done)) { done(); }
```
bendiy: What is this?

test/database/dblib.js
```diff
@@ -74,19 +74,19 @@
             } else {
               assert.equal(res.rowCount, 1);
             }
-            if (_.isFunction(done)) { done(); }
+            sp();
```
gilmoskowitz: `done()` doesn't get called at the end of `createUser()` so it's not going to be synchronized with a test that needs to create a user. Same with`deketeUser()`.

bendiy: It's there in `sr`:
```
 //          assert.isNull(res);
             if (_.isFunction(done)) { done(); }		             if (_.isFunction(done)) { done(); }
```

test/database/dblib.js
```diff
@@ -74,19 +74,19 @@
...
                           "\" with password '" + creds.password + "';";
           datasource.query(passwdSql, adminCred, function (err, res) {
 //          assert.isNull(err, 'expect no error changing the user password');
 //          assert.isNull(res);
-            if (_.isFunction(done)) { done(); }
+            sr();
           });
         },
-        sr = function (done) {
+        sr = function () {
```
bendiy: `done()` is called below, just not shown in this diff.

test/database/triggers/crmacct.js
```diff
@@ -16,40 +16,59 @@ var _       = require('underscore'),
...
+              "delete from crmacct  where crmacct_number = $1;"
+            ];
+            deleteSql.forEach(function (sql, deleteIndex) {
+              var cred = _.extend({}, adminCred, { parameters: [ acct ] });
+              datasource.query(sql, cred, function (err, res) {
+                assert.isNull(err);
+                assert.isNotNull(res);
+
+                if (index === accounts.length - 1 && deleteIndex === deleteSql.length - 1) {
+                  done();
```
bendiy: Call's `done()` when everything is is finished.

test/database/dblib.js
```diff
@@ -95,36 +95,34 @@
...
+        dxtu = function () {
           var sql = 'delete from xt.usrlite where usr_username = $1;';
           datasource.query(sql, context, function (err, res) {
             assert.isNull(err, 'delete ' + username + ' xt.userlite should work');
-            if (_.isFunction(done)) { done(); }
+            du();
           });
         },
-        du = function (done) {
+        du = function () {
```
bendiy: Calls `done()` below, but not shown in this diff.

node-datasource/lib/ext/datasource.js
```diff
@@ -16,7 +16,31 @@ Backbone:true, _:true, X:true, __dirname:true, exports:true, module: true */
...
+       * connection, the `search_path` will not be set the next time the pool's
+       * conneciton is used for a query.
+       *
+       * Because of this, on any emitted errors below in `connected()`, we
+       * destroy that client from the pool.
+       */
+      destroyClientFromPool: function (client) {
+        client.destroying = true;
+        X.pg.pools.all[JSON.stringify(this.creds)].destroy(client);
+      },
```
bendiy: That is correct. I had considered only destroying on the first query for a connection, but it all depends on the query that first calls some plv8 code. The error could happen on any query during the connection. 

Updated the comment to reflect that.

node-datasource/lib/ext/datasource.js
```diff
@@ -142,6 +166,12 @@ Backbone:true, _:true, X:true, __dirname:true, exports:true, module: true */
           return callback(err);
         }
 
+        // If we got a client that was flagged to be destroyed, try again.
+        if (client.destroying) {
+          this.query(query, options, callback);
+          return;
+        }
```
bendiy: The flow is:
1. `query: function (query, options, callback)` gets a client from the pool and calls:
2. `connected: function (query, options, callback, err, client, done)` executes the query against the database, handles the response and errors. Actual execution is done below at:
```
        if (options.parameters) {
          client.query(query, options.parameters, queryCallback);
        } else {
          client.query(query, queryCallback);
        }
```

Due to the timing/async nature of the calling code, the time it takes for `X.pg.pools.all[JSON.stringify(this.creds)].destroy(client);` to actually remove the client connection from the pool is slower than the next incoming query gets a client from the pool. In my testing, some queries were getting the same client I had just requested to be destroyed. By flagging it first, then calling `destroy(client)`, I can ensure another incoming query doesn't use it while it's in the process of being destroyed.

PR #3121, opened 2017-08-07
issue #30679 Convert cntctsel to temp table so it works with multiple sessions

ChristopherCotnoir:


ChristopherCotnoir:
Associated with qt-client#1726.

PR #3122, opened 2017-08-07
Issue #29222: AP Open FX rate discrepancy from GL ledger posting.

andersdd:


harithamekala:
Unable to do buildapp on Pull request masterref database due to issue #30721

PR #3136, opened 2017-08-23
issue #30607 Fix handling of checkitem_curr_rate

ChristopherCotnoir:


harithamekala:
Performed UAT: Issue is fixed. Check is voided successfully with no error. Please refer to the attached screenshot ‘Void.png’ 

void
and ‘Void1.png’
void

PR #3140, opened 2017-08-29
Expose XM.SalesOrder.addPayment dispatch service to the REST API.

bendiy:
Documentation:
https://github.com/xtuple/xtuple/wiki/REST-API-Examples#example-requests
https://github.com/xtuple/xtuple/wiki/REST-API-Example:-Associate-Customer-Deposit-with-Sales-Order

@gilmoskowitz For this to work in 4.11.0, I would need to add this to the xdruple-extension as well. Should I do that?

### DIFF
--- a/database/source/xm/javascript/sales_order.sql
+++ b/database/source/xm/javascript/sales_order.sql
@@ -101,6 +101,45 @@ select xt.install_js('XM','SalesOrder','xtuple', $$
       paid: !!payment
     };
   };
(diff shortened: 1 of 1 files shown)


gilmoskowitz:
@bendiy If we need this to work with 4.11.0 servers then yes, add it to the xdruple-extension along with an `@deprecated` or equivalent comment explaining that the code is duplicated, where the "master" is (xtuple repo) and why.

PR #3142, opened 2017-08-31
issue #30867 Handle BETWEEN properly

ChristopherCotnoir:


harithamekala:
Performed UAT: Changed the system date to 31/08/2017 and observed that transactions are posted successfully. Issue is fixed.

PR #3144, opened 2017-09-08
issue #30903 Don't alter pkghead while querying pkghead, rewrite packageisenabled …

ChristopherCotnoir:
…to work properly for schemas with lots of table inheritance not related to packages


rkannadevara:
Performed UAT: Issue is fixed. no issue is observed while enabling/disabling all packages. Please refer the below attached screenshot.

listpackages

PR #3145, opened 2017-09-11
issue #30822 Take returns into account when determining receivable amount

ChristopherCotnoir:


harithamekala:
Unable to verify the fix due to issue #30932

PR #3146, opened 2017-09-11
Issue #29880: Document/Image mass import helper views/functions

andersdd:


foundation-database/public/functions/createimage.sql
```diff
@@ -0,0 +1,16 @@
...
+                                        pDescription text,
+                                        pImageData TEXT)
+  RETURNS integer AS $$
+-- Copyright (c) 1999-2017 by OpenMFG LLC, d/b/a xTuple. 
+-- See www.xtuple.com/CPAL for the full text of the software license.
+DECLARE
+  _id integer;
+BEGIN
+  INSERT INTO image (image_id, image_name, image_descrip, image_data) 
+             VALUES (nextval('image_image_id_seq'), pTitle, pDescription, pImageData)
```
gilmoskowitz: Just let `image_id` default, either by leaving it out entirely or passing `DEFAULT` instead of `nextval()`.

foundation-database/public/views/document_import.sql
```diff
@@ -0,0 +1,107 @@
+DROP FUNCTION IF EXISTS public.importdocument(documentimport) cascade;
```
gilmoskowitz: This line may be responsible for the CI failure. The `documentimport` view, therefore the type of the same name, don't exist yet. This drop function must follow the view creation.

andersdd: Nope different problem - it appears if you do not explicitly define a view schema it defaults into `xt`.

andersdd:
@gilmoskowitz Not sure why CI is failing here.  I have run the sql script on several Dbs with and without existing views, without the script causing errors.

PR #3147, opened 2017-09-12
Issue #29316: Add trigger functions to remove document/image assignme…

andersdd:
…nts and prevent orphaned records

I did _not_ add the reverse checks (assignments deleted but not the image/document) as you can pick up an image and assign to different records, and we want to be able to do the same for documents/files in the future.


harithamekala:
Unable to verify the fix due to issue #30932

foundation-database/public/tables/imageass.sql
```diff
@@ -0,0 +1,26 @@
...
+SELECT
+  xt.add_constraint('imageass', 'imageass_pkey', 'PRIMARY KEY (imageass_id)', 'public'),
+  xt.add_constraint('imageass', 'imageass_imageass_purpose_check', $$CHECK (imageass_purpose IN ('I','E','M','P','A','C','D','S')$$, 'public');
+
+
+DELETE FROM imageass
+WHERE imageass_image_id IN (SELECT imageass_image_id
+                            FROM imageass 
+                            LEFT OUTER JOIN image ON (imageass_image_id=image_id)
+                            WHERE image_id IS NULL);
```
gilmoskowitz: There should be a more straightforward way to phrase this:
```SQL
WHERE NOT EXISTS(SELECT 1
                   FROM image
                  WHERE image_id = imageass_image_id);
```

harithamekala:
Issue still exists,Documents attached to incident are not being deleted from ‘docass’ table when incident is cancelled during its creation.Please refer to the attached screenshot ‘DocassTable.png’
docasstable

andersdd:
@harithamekala Please retest.  Additional logic added to clean up upon deletion of incidents.

PR #3154, opened 2017-09-18
issue #29248 Allow assigning status to an imported revision

ChristopherCotnoir:


ChristopherCotnoir:
Requires xtuple/private-extensions#957.

foundation-database/api/views/bom.sql
```diff
@@ -9,11 +9,13 @@ SELECT dropIfExists('VIEW', 'bom', 'api');
     bomhead_docnum AS document_number,
     bomhead_revisiondate AS revision_date,
     bomhead_batchsize AS batch_size,
-    bomhead_requiredqtyper AS total_qty_per
+    bomhead_requiredqtyper AS total_qty_per,
+    rev_status AS revision_status
   FROM
-    bomhead, item
+    bomhead, item, rev
```
gilmoskowitz: The view cannot do this because PostBooks has bills of materials but does not have the `rev` table.

foundation-database/public/functions/savebomhead.sql
```diff
@@ -1,4 +1,4 @@
-CREATE OR REPLACE FUNCTION saveBomHead(integer,text,date,text,numeric,numeric)
+CREATE OR REPLACE FUNCTION saveBomHead(integer,text,date,text,numeric,numeric,text)
```
gilmoskowitz: Maintain backwards compatibility. Either define 2 versions of the function or make the new last argument default to a reasonable value (`NULL` is reasonable).

PR #3159, opened 2017-09-22
Fixes Issue #30663 Call done(err) instead, which destroy's the client and removed it from the pool.

bendiy:
The last fix disconnected the client and removed it from the pool by calling `X.pg.pools.all[JSON.stringify(this.creds)].destroy(client);`. However, the `done()` callback was still called below that, which reconnected a new client and threw off the pool count. Line 272 here:
https://github.com/xtuple/xtuple/blob/6a2718719596b1e03ac83f47d07ac9d51f56de94/node-datasource/lib/ext/datasource.js#L227-L272

That lead to the active pg connections building until the pg server hit 100 connections and stopped accepting any more.

By calling `done(err)` instead of `X.pg.pools.all[JSON.stringify(this.creds)].destroy(client);`, `destroy()` get called anyways, which was the original goal of the last commit. The call happens here:
https://github.com/brianc/node-postgres/blob/v0.14.1/lib/pool.js#L109-L113

### DIFF
--- a/node-datasource/lib/ext/datasource.js
+++ b/node-datasource/lib/ext/datasource.js
@@ -20,28 +20,6 @@ Backbone:true, _:true, X:true, __dirname:true, exports:true, module: true */
 
       creds : {},
 
-      /**
-       * The package, `node-postgres`, we use, exposed as `X.pg` maintains a
-       * connection pool to the database server, defaulting to 10 connections.
-       *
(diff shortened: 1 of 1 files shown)


gilmoskowitz:
jenkins - retest this please

bendiy:
@gilmoskowitz CI's passing now.

PR #3161, opened 2017-09-25
make translations in database optional

gilmoskowitz:


scripts/release_build.sh
```diff
@@ -20,11 +21,11 @@ XTUPLEDIR=$(pwd)
 #                     |       |        |     path for build_app -e, relative to repo root
 #  module             tag     edition  build source
 declare -a CONFIG=(\
-  "xtuple             ARGS    skip     skip  not-needed"                                  \
-  "private-extensions ARGS    skip     skip  not-needed"                                  \
-  "qt-client          ARGS    skip     skip  not-needed"                                  \
+  "xtuple             skip    skip     skip  not-needed"                                  \
+  "private-extensions skip    skip     skip  not-needed"                                  \
+  "qt-client          skip    skip     skip  not-needed"                                  \
```
ChristopherCotnoir: This will cause a failure even in no translations mode because the updater is still being built and run, and that causes a problem if qt-client doesn't already exist. private-extensions already exists if called from nightlypackage.sh in xt-qtdev-tools, but otherwise skipping can cause a failure.

gilmoskowitz: It continues to work because of https://github.com/gilmoskowitz/xtuple/blob/1c29f152e787719752f30134b5f53921d9465eb0/scripts/release_build.sh#L106-L109

PR #3170, opened 2017-10-02
issue #31053 - added upgrade test, standardized upgrade internal dir names

gilmoskowitz:
tested with 4.4.0, 4.5.0, 4.6.0, 4.8.1, 4.9.0 -> 4.11.1

### DIFF
--- a/scripts/UpgradeTest.sh
+++ b/scripts/UpgradeTest.sh
@@ -53,25 +53,28 @@ applyUpdate() {
   $RUN $UPDATER -h $PGHOST -p $PGPORT -d $DB -U $PGUSER -passwd=$PGPASS -D -autorun -f $FILE
 }
 
+getValueArg() {
+  local KEY="$1"
+  local VALUE="$2"
+  if [ -n "$KEY" -a -n "$VALUE" ] ; then
+    eval $KEY="$VALUE"
+    return 0
+  fi
+  return 1
+}
+
+# -h is handled as either --host if it has an arg or --help if not
 while [ $# -gt 0 ] ; do
   case "$1" in
-    --help)         usage              ; exit 0 ;;
-    -h|--host)
-      if [ -n "$2" ] ; then
-        PGHOST=$2
-        shift
-      else
-        usage
-        exit 0
-      fi
-      ;;
-    -p|--port)      PGPORT=$2          ; shift  ;;
-    -U|--user)      PGUSER=$2          ; shift  ;;
-    --pass*)        PGPASSWORD=$2      ; shift  ;;
-    -u|--updater)   UPDATER="$2"       ; shift  ;;
-    -n|--no-run)    RUN=echo           ; shift  ;;
-    -P|--public)    PUBLICDIR=$2       ; shift  ;;
-    *)              usage              ; exit 1 ;;
+    --help)       usage ; exit 0 ;;
+    -n|--no-run)  RUN=echo       ;;
+    -h|--host)    if getValueArg PGHOST     $2 ; then shift ; else usage ; exit 0 ; fi ;;
+    -p|--port)    if getValueArg PGPORT     $2 ; then shift ; else usage ; exit 1 ; fi ;;
+    -U|--user)    if getValueArg PGUSER     $2 ; then shift ; else usage ; exit 1 ; fi ;;
+    --pass*)      if getValueArg PGPASSWORD $2 ; then shift ; else usage ; exit 1 ; fi ;;
+    -u|--updater) if getValueArg UPDATER    $2 ; then shift ; else usage ; exit 1 ; fi ;;
+    -P|--public)  if getValueArg PUBLICDIR  $2 ; then shift ; else usage ; exit 1 ; fi ;;
+    *)            usage ; exit 1 ;;
   esac
   shift
 done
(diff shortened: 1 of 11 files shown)


scripts/UpgradeTest.sh
```diff
@@ -0,0 +1,135 @@
...
+  case "$1" in
+    --help)         usage              ; exit 0 ;;
+    -h|--host)
+      if [ -n "$2" ] ; then
+        PGHOST=$2
+        shift
+      else
+        usage
+        exit 0
+      fi
```
ChristopherCotnoir: Why this check only for host and not for the other args?

scripts/UpgradeTest.sh
```diff
@@ -0,0 +1,135 @@
...
+      ;;
+    -p|--port)      PGPORT=$2          ; shift  ;;
+    -U|--user)      PGUSER=$2          ; shift  ;;
+    --pass*)        PGPASSWORD=$2      ; shift  ;;
+    -u|--updater)   UPDATER="$2"       ; shift  ;;
+    -n|--no-run)    RUN=echo           ; shift  ;;
```
ChristopherCotnoir: Should not shift here.

scripts/UpgradeTest.sh
```diff
@@ -0,0 +1,135 @@
...
+  shift
+done
+
+ls ${PUBLICDIR}/[1-9].[0-9].[0-9]/*.backup ${PUBLICDIR}/[1-9].[0-9][0-9].[0-9]/*.backup
+for FILE in $(ls ${PUBLICDIR}/[1-9].[0-9].[0-9]/*.backup ${PUBLICDIR}/[1-9].[0-9][0-9].[0-9]/*.backup) ; do
+    STARTVER=$(basename $(dirname $FILE))
+    if [ -e ${PUBLICDIR}/$STARTVER/masterref*${STARTVER}*.backup ] ; then
+      DB=$(basename $FILE -${STARTVER}.backup)_to${DESTVER}
+    else
+      DB=$(basename $FILE -${STARTVER/./}.backup)_to${DESTVER}
```
ChristopherCotnoir: What is this for? I don't think it does what it is supposed to do, whatever that is.

scripts/UpgradeTest.sh
```diff
@@ -0,0 +1,135 @@
...
+    $RUN psql -t -d $DB -c "SELECT COUNT(*) = 0 OR BOOL_AND(pkghead_version = '$DESTVER') \
+                           FROM pkghead WHERE pkghead_name in ('xtcore', 'xtmfg', 'xwd');" | tee $TMPDIR/$PROG.$$
+    [ -n "$RUN" -o "$(tr -d [:space:] < $TMPDIR/$PROG.$$)" = 't' ] || logErr Core extensions in $DB do not match $DESTVER
+
+    echo =========================================================================
+done
+
+applyUpdate mfg_quickstart_to${DESTVER}       $SCRIPTDIR/manufacturing-upgrade-*.gz || logErr error reapplying manufacturing upgrade
+applyUpdate postbooks_quickstart_to${DESTVER} $SCRIPTDIR/postbooks-upgrade-*.gz     || logErr error reapplying postbooks upgrade
+applyUpdate distquickstart_to${DESTVER}       $SCRIPTDIR/add-manufacturing-to-distribution-*.gz || logErr error upgrading distribution to enterprise
```
ChristopherCotnoir: Why not move these in the loop above? I assume these are testing idempotency? Could simply rerun applyUpdate at the bottom of the loop.

scripts/xml/shared.xml
```diff
@@ -16,13 +16,6 @@
   </prerequisite>
 
   <prerequisite type = "query"
-               name = "Checking if plv8 extension is installed" >
-    <query>SELECT exists(SELECT 1 FROM pg_catalog.pg_extension WHERE extname = 'plv8');</query>
-    <message>The PostgreSQL extension, plv8, must be installed on your database. See: https://github.com/xtuple/xtuple/wiki/Installing-PLv8
-    </message>
-  </prerequisite>
-
-  <prerequisite type = "query"
```
ChristopherCotnoir: Are we sure this is safe? I thought plv8 was required because of xt.js_init().

gilmoskowitz: This is safe because of https://github.com/xtuple/xtuple/blob/4_11_x/lib/orm/source/create_plv8.sql#L5 and https://github.com/xtuple/xtuple/blob/4_11_x/foundation-database/manifest.js#L21

scripts/UpgradeTest.sh
```diff
@@ -0,0 +1,133 @@
...
+    [ -n "$RUN" -o "$(tr -d [:space:] < $TMPDIR/$PROG.$$)" = 't' ] || logErr ServerVersion in $DB does not match $DESTVER
+    $RUN psql -t -d $DB -c "SELECT COUNT(*) = 0 OR BOOL_AND(pkghead_version = '$DESTVER') \
+                           FROM pkghead WHERE pkghead_name in ('xtcore', 'xtmfg', 'xwd');" | tee $TMPDIR/$PROG.$$
+    [ -n "$RUN" -o "$(tr -d [:space:] < $TMPDIR/$PROG.$$)" = 't' ] || logErr Core extensions in $DB do not match $DESTVER
+    applyUpdate $DB $GZ || logErr error reapplying upgrade to $DESTVER for $DB = $FILE using $GZ
+
+    echo =========================================================================
+done
+
+applyUpdate distquickstart_to${DESTVER}       $SCRIPTDIR/add-manufacturing-to-distribution-*.gz || logErr error upgrading distribution to enterprise
```
ChristopherCotnoir: This should be in the loop too and applied to all distribution backups rather than just applying to whatever the last backup that ended up with this database name was.

PR #3174, opened 2017-10-05
Issue #31105: Vendor default tax type on misc distributions

andersdd:


foundation-database/public/tables/vendinfo.sql
```diff
@@ -44,6 +44,7 @@ SELECT
   xt.add_column('vendinfo', 'vend_accnt_id',         'INTEGER', NULL,                            'public'),
   xt.add_column('vendinfo', 'vend_expcat_id',        'INTEGER', 'DEFAULT (-1)',                  'public'),
   xt.add_column('vendinfo', 'vend_tax_id',           'INTEGER', 'DEFAULT (-1)',                  'public'),
+  xt.add_column('vendinfo', 'vend_taxtype_id',       'INTEGER', 'DEFAULT (-1)',                  'public'),
```
gilmoskowitz: This (really "all three of these") should be a real foreign key, especially because it's new. This will ripple to the C++ save code.

andersdd: Corrected foreign key relationships for all 4 fields.  Flushed through to Vendor and Voucher Misc Distribution screen, but I have not made changes to the vodist processing as this affects subsequent function processing.  Will add to the list though.

PR #3175, opened 2017-10-05
issue #30893 - site api change

tomdatkins:


tomdatkins:
Changed the tax_zone to not use -1 on the insert.

foundation-database/api/views/site.sql
```diff
@@ -1,13 +1,12 @@
-  --Site (aka Warehouse) View
+-- View: api.site
 
-  SELECT dropIfExists('VIEW', 'site', 'api');
```
gilmoskowitz: Let this script drop the view unless there's a compelling reason not to. If that's the case, please document why we should not drop it.

foundation-database/api/views/site.sql
```diff
@@ -16,632 +15,361 @@
...
+            ELSE NULL::integer
+        END, COALESCE(getsitetypeid(new.type), '-1'::integer), COALESCE(new.scheduling_sequence, 0));
+
+-- Rule: "_UPDATE" ON api.site
+
+-- DROP RULE "_UPDATE" ON api.site;
+
+CREATE OR REPLACE RULE "_UPDATE" AS
+    ON UPDATE TO api.site DO INSTEAD  UPDATE whsinfo SET warehous_descrip = new.description, warehous_fob =
+        CASE
```
gilmoskowitz: Rearranging the code so the `column =` is separated from the value calculation makes the code much harder to verify.

gilmoskowitz:
@tomdatkins The initial comment indicates this pull request is a one line change. Somehow this has blown up into 37 files with a code churn of about 1600 lines. Please submit just the one little piece that you want to fix.

PR #3178, opened 2017-10-09
Issue #30799: Improve relationship between voided voucher and CM tax

andersdd:
Still requires Tax assignment to be correctly configured otherwise GL to tax history discrepancies can occur, but makes assignment a bit more robust in the case of voided vouchers

### DIFF
--- a/foundation-database/public/functions/updatememotax.sql
+++ b/foundation-database/public/functions/updatememotax.sql
@@ -1,4 +1,4 @@
-DROP FUNCTION IF EXISTS updatememotax(text, text, integer, integer, date, integer, numeric);
+DROP FUNCTION IF EXISTS updatememotax(text, text, integer, integer, date, integer, numeric, numeric);
 
 CREATE OR REPLACE FUNCTION updatememotax(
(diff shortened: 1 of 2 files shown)


rkannadevara:
Performed UAT: Issue is fixed. Observed that tax history for credit memo of voided voucher is displayed in ‘Tax History’ screen when ‘Automatically Apply to AP/AR Memos’ is enabled for ‘Adjustment’ tax type. Please refer the below attached screenshot:

voidvoucher

foundation-database/public/functions/updatememotax.sql
```diff
@@ -1,4 +1,4 @@
-DROP FUNCTION IF EXISTS updatememotax(text, text, integer, integer, date, integer, numeric);
+DROP FUNCTION IF EXISTS updatememotax(text, text, integer, integer, date, integer, numeric, numeric);
```
gilmoskowitz: We still need to drop the 7-arg version if it exists. We don't know from how far back people are starting their upgrade.

PR #3184, opened 2017-10-16
Merge report translations into regular translation files, remove qt-c…

ChristopherCotnoir:
…lient translations


gilmoskowitz:
Please remove qt-client/share/dict/reports.base.ts if it's no longer needed.

PR #3185, opened 2017-10-18
issue #31168 Move filter condition to where clause

ChristopherCotnoir:


rkannadevara:
Performed UAT: Issue is fixed. Observed that duplicate lines are not displayed in ‘Detailed Item Costs’ screen. Please refer the below attached screenshot:

costingdetail

PR #3186, opened 2017-10-18
issue #30592 Link cash receipts with ccpay

ChristopherCotnoir:


rkannadevara:
Performed UAT: Issue is fixed. Observed that cash receipt is not saved and no error dialog is displayed while saving a cash receipt when credit card payment is declined. Please refer the below attached screenshot:

cashreceiptslist

foundation-database/public/patches/migrateccpay.sql
```diff
@@ -0,0 +1,188 @@
...
+                          AND ccpay_order_number=_r.docnumber
+                          AND ccpay_cust_id=_r.cashrcpt_cust_id
+                        ORDER BY ccpay_transaction_datetime, ccpay_id);
+
+      FOREACH _cashrcpt IN ARRAY _cashrcpts
+      LOOP
+        SELECT cashrcpt_amount, cashrcpt_curr_id
+          INTO _amount, _currid
+          FROM cashrcpt
+         WHERE cashrcpt_id=_cashrcpt;
```
gilmoskowitz: Either fold the `FOREACH` and `SELECT` together so there's one query instead of length(_cashrcpts) queries, or add two more `array_agg` calls in the query that populates `_r` to create parallel arrays. Then this loop over the cash receipts could walk through the arrays using a shared index variable to get the desired data.

foundation-database/public/patches/migrateccpay.sql
```diff
@@ -0,0 +1,188 @@
...
+
+              EXIT;
+            END IF;
+          END LOOP;
+        END IF;
+      END LOOP;
+
+      FOREACH _cashrcpt IN ARRAY _cashrcpts
+      LOOP
+        IF (_cashrcpt NOT IN (SELECT UNNEST(_foundcashrcpts))) THEN
```
gilmoskowitz: Consider using [array_remove](https://www.postgresql.org/docs/9.3/static/functions-array.html) to shrink `_cashrcpts` on the fly instead of collecting `_foundcashrcpts`.

foundation-database/public/patches/migrateccpay.sql
```diff
@@ -0,0 +1,188 @@
...
+
+              EXIT;
+            END IF;
+          END LOOP;
+        END IF;
+      END LOOP;
+    END IF;
+  END LOOP;
+
+END;
```
gilmoskowitz: There are four sets of nested loops here corresponding to the amount of information we have available to match cash receipts to ccpay records. They can be combined into one set if enough information is collected up front and the loops `SELECT ... ORDER BY theAmountOfAvailableInformation DESC`. Alternatively, put all 4 cases together and use either `IF ... ELSIF ...` or [EXIT label](https://www.postgresql.org/docs/9.3/static/plpgsql-control-structures.html#PLPGSQL-CONTROL-STRUCTURES-LOOPS).

foundation-database/public/patches/migrateccpay.sql
ChristopherCotnoir: It's dangerous to modify an array while looping over it.

xtuple-ci-user:
Build finished.

gilmoskowitz:
+1 but let's hold off merging for now. One of our partners is going to test this on a copy of a customer database to see if this works on a larger scale.

bendiy:
@ChristopherCotnoir 
Take a look at:
https://github.com/xtuple/xtuple/blob/387fb8f9bca1baf769a8ee6d373dc2e06a64a810/foundation-database/public/functions/postcashreceipt.sql#L100-L121
We might be able to clean that up now.

Possibly related:
https://forums.xtuple.com/t/new-user-cc-chg-post-error/5872/5

I have some code in `payment-gateways` that will need to be updated as well.

PR #3187, opened 2017-10-19
issue #29744 Allow user to enter date for deposits and journal series in general

ChristopherCotnoir:


ChristopherCotnoir:
Associated with xtuple/qt-client#1803.

PR #3188, opened 2017-10-23
issue #30908 Don't post lower level costs

ChristopherCotnoir:


mccjim:
Made this change in a test db and it appears to have resolved the issue.

rz_090717 on the support server.

foundation-database/public/functions/postproduction.sql
```diff
@@ -204,6 +204,7 @@ BEGIN
     (costelem_id = itemcost_costelem_id) AND
     (itemcost_item_id = itemsite_item_id) AND
     (itemsite_costcat_id = costcat_id) AND
+    (NOT itemcost_lowlevel) AND
```
gilmoskowitz: This proposed change is very simple. However, it does not explain how the low level costs were handled properly in earlier versions. Something else must have changed between 4.9.x and 4.10.1 to cause thiis bug.

PR #3189, opened 2017-10-24
issue #31245 Better preserve history

ChristopherCotnoir:


ChristopherCotnoir:
Associated with xtuple/qt-client#1809.

PR #3190, opened 2017-10-30
issue #31229 Use LEFT OUTER JOIN for shipform, improve readability

ChristopherCotnoir:


rkannadevara:
Performed UAT: Issue is fixed. Observed that sales order is displayed in 'Sales Order' API view with 'shipping_form' column as NULL when sales order is created with no shipping form. Please refer the below attached screenshot:

shipform

PR #3191, opened 2017-10-30
adding service script template/examples

xtpclark:
These should live in node-datasource.


node-datasource/service-scripts/debian-init
```diff
@@ -0,0 +1,162 @@
+#!/bin/sh
+
+APP_DIR="/opt/xtuple/${MWCVERSION}/${MWCNAME}/xtuple/node-datasource"
+NODE_APP="main.js"
+CONFIG_FILE="/etc/xtuple/${MWCVERSION}/${MWCNAME}/config.js"
```
gilmoskowitz: Please don't use the prefix `MWC`. `XT` or `XTUPLE` would be more appropriate.

node-datasource/service-scripts/debian-init
```diff
@@ -0,0 +1,162 @@
...
+    [ -f "$PID_FILE" ]
+}
+
+get_pid() {
+    echo "$(cat "$PID_FILE")"
+}
+
+is_running() {
+    PID=$(get_pid)
+    ! [ -z "$(ps aux | awk '{print $2}' | grep "^$PID$")" ]
```
gilmoskowitz: Simplify?
```
ps -p "$(get_pid)"
```

node-datasource/service-scripts/debian-init
```diff
@@ -0,0 +1,162 @@
...
+}
+
+start_it() {
+    mkdir -p "$PID_DIR"
+    mkdir -p "$LOG_DIR"
+
+    echo "Starting xTuple ERP..."
+    $NODE_EXEC "$APP_DIR/$NODE_APP" -c $CONFIG_FILE 1>"$LOG_FILE" 2>&1 &
+    echo $! > "$PID_FILE"
+    echo "xTuple ERP started with pid $!"
```
gilmoskowitz: We should be more precise here and below. We're starting a specific component of the ERP system - the REST API server.

node-datasource/service-scripts/debian-init
```diff
@@ -0,0 +1,162 @@
...
+                echo "xTuple ERP stopped"
+            else
+                exit 1
+            fi
+        fi
+    else
+        echo "xTuple ERP already stopped, pid file does not exist"
+        exit 1
+    fi
+}
```
gilmoskowitz: This can be simplified if we stop worrying about the details. The existence of the pid file probably doesn't matter if the process isn't running; just remove it after the process has been stopped.

The goal of FORCE_OP should be to kill the process even if it doesn't want to die:
```bash
echo "Stopping"
stop_process
sleep 1
if $FORCE_OP && is_running ; then
  kill -9 $(get_pid)
  sleep 1
  if is_running ; then
    echo "It just won't go away:-("
    exit 1
  fi
fi
remove_pid_file
echo "Stopped"
```

gilmoskowitz:
closed after discussing with @xtpclark - this duplicates code in [xTAU](https://github.com/xtuple/xtuple-admin-utility) and isn't needed here.

PR #3192, opened 2017-11-02
issue #31304 Properly exclude future payments

ChristopherCotnoir:


ChristopherCotnoir:
Associated with xtuple/qt-client#1818

PR #3194, opened 2017-11-10
issue #31441 Add and populate file_mime_type column

ChristopherCotnoir:


foundation-database/public/functions/createfile.sql
```diff
@@ -1,12 +1,9 @@
-create or replace function createFile(text, text, bytea) returns integer as $$
+create or replace function createFile(pTitle TEXT, pDescription TEXT, pStream BYTEA, pMimeType TEXT) returns integer as $$
```
bendiy: `pMimeType` needs  a default `'application/octet-stream'` when it's `NULL` or a `COALESCE(pMimeType, 'application/octet-stream')` on the `INSERT` below. I just crashed the Qt Client trying to add a file through DocAttach.

```
Tue Nov 14 13:27:58 2017 ERROR:  null value in column "file_mime_type" violates not-null constraint
DETAIL:  Failing row contains (3, Incidents.png, \211PNG\015\012\032\012\000\000\000\015IHDR\000\000\005\201\000\..., Incidents.png, null, 7f8d83b6-fb26-4b5e-9fd8-d3d27a98737a).
CONTEXT:  SQL statement "insert into file (file_id, file_title, file_descrip, file_stream, file_mime_type) values (_id, pTitle, pDescription, pStream, pMimeType)"
PL/pgSQL function createfile(text,text,bytea,text) line 6 at SQL statement
(23502) QPSQL: Unable to create query
INSERT INTO url ( url_source, url_source_id, url_title, url_url, url_stream ) VALUES ( 'CRMA', 1, 'Incidents.png', 'Incidents.png',
```

ChristopherCotnoir: I always forget that NOT NULL DEFAULT columns don't behave the way I would like on explicit NULL insertions. Fixed.

bendiy: The `COALESCE` is working for me. Still have not fully reviewed this.

foundation-database/public/views/url.sql
```diff
@@ -84,7 +86,8 @@ create or replace rule "_UPDATE_FILE" as on update to url
 
 update file set
   file_title = new.url_title,
-  file_stream = new.url_stream
+  file_stream = new.url_stream,
+  file_mime_type = new.url_mime_type
```
gilmoskowitz: This also needs a `COALESCE`. Alternatively, put that in a `before` trigger so it's handled regardless of what body of code updates the `file` table.

PR #3196, opened 2017-11-16
issue #29638 - New deleteCntct function

ChristopherCotnoir:


ChristopherCotnoir:
Associated with xtuple/qt-client#1830.

PR #3197, opened 2017-11-17
issue #19108 Add script fetch metasql

ChristopherCotnoir:


ChristopherCotnoir:
Associated with xtuple/qt-client#1827.

PR #3206, opened 2017-12-15
issue #31730 - handle vendor group instead of cluster

gilmoskowitz:
also reduced redundancy to focus on the important
differences in the various parts of the union

### DIFF
--- a/foundation-database/public/tables/metasql/vendorAPHistory-detail.mql
+++ b/foundation-database/public/tables/metasql/vendorAPHistory-detail.mql
@@ -1,23 +1,20 @@
 -- Group: vendorAPHistory
 -- Name: detail
 -- Notes: used by dspVendorAPHistory
--- Copyright (c) 1999-2014 by OpenMFG LLC, d/b/a xTuple.
--- See www.xtuple.com/CPAL for the full text of the software license.
+--        Copyright (c) 1999-2017 by OpenMFG LLC, d/b/a xTuple.
+--        See www.xtuple.com/CPAL for the full text of the software license.
 
+WITH mainQuery AS (
 SELECT apopen_id, -1 AS applyid, 0 AS type,
        apopen_docdate AS sortdate, apopen_docnumber AS sortnumber,
        apopen_docnumber AS docnumber,
        formatBoolYN(apopen_open) AS f_open,
-       CASE WHEN (apopen_doctype='V') THEN <? value("voucher") ?>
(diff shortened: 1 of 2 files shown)


foundation-database/public/tables/metasql/vendorAPHistory-detail.mql
```diff
@@ -1,94 +1,116 @@
...
+                             JOIN vendtype ON vend_vendtype_id = vendtype_id
+                            WHERE vendtype_code ~* <? value('vendtype_pattern') ?>)
+  <? endif ?>
+)
+SELECT mainQuery.*,
+       CASE doctype WHEN 'V' THEN <? value('voucher') ?>
+                    WHEN 'C' THEN <? value('creditMemo') ?>
+                    WHEN 'D' THEN <? value('debitMemo') ?>
+                    WHEN 'K' THEN <? value('check') ?>
+                    ELSE doctype
```
ChristopherCotnoir: Shouldn't this be <? value("other") ?> as it was before?

foundation-database/public/tables/metasql/vendorAPHistory-detail.mql
```diff
@@ -1,94 +1,116 @@
...
+    AND apopen_vend_id IN (SELECT vend_id FROM vendinfo
+                            WHERE vend_vendtype_id = <? value('vendtype_id') ?>)
+  <? elseif exists('vendtype_pattern') ?>
+    AND apopen_vend_id IN (SELECT vend_id
+                             FROM vendinfo
+                             JOIN vendtype ON vend_vendtype_id = vendtype_id
+                            WHERE vendtype_code ~* <? value('vendtype_pattern') ?>)
+  <? endif ?>
+)
+SELECT mainQuery.*,
```
ChristopherCotnoir: I thought we were avoiding SELECT *?

foundation-database/public/tables/metasql/vendorAPHistory-detail.mql
gilmoskowitz: `<? value('other') ?>'? That makes debugging really hard because it removes an essential clue about where the bad records are coming from.

foundation-database/public/tables/metasql/vendorAPHistory-detail.mql
gilmoskowitz: We're avoiding `select *` when tables are joined. In this case, any problem with duplicate columns would have been found and reported within `mainQuery`.

PR #3208, opened 2017-12-27
issue #31733 - don't show closed items by default

gilmoskowitz:
also add/fix some tests


foundation-database/public/tables/metasql/orderActivityByProject-detail.mql
```diff
@@ -760,16 +780,19 @@ SELECT invchead_id AS id,
 <? endif ?>
     JOIN invcitem ON (invcitem_invchead_id = invchead_id)
  WHERE (invchead_prj_id = <? value("prj_id") ?>)
+<? if not exists('showCompleted') ?>
+   AND invchead_posted
```
ChristopherCotnoir: This seems backwards. Shouldn't this be AND NOT invchead_posted, here and below? I assume posted=completed.

PR #3205, opened 2017-12-08
CRM Enhancement project [xTuple v5.0 part #1]

andersdd:
Significant structural changes in the CRM system.
Added concept of a CRM Role with associated setup up options.
`crmacct` and `cntct` now have a many-to-many relationship saved in `crmacctcntctass` table.  Old CRM contacts (1&2) removed from table.  CRM account removed from cntct table.
Multiple addresses available on CRM account table.
Relationship between crmacct and child entities (customer/vendor) now reversed and set up as actual foreign key relations on the child tables (i.e. cust_crmacct_id).  Old crm account relationship columns removed.
Contact phones (phone1,2, fax) removed from cntct table and moved to a cntctphone table which allows any number of phone numbers and types to be saved.
Data migration script should move old structure data into new structures.
Both standard and commercial databases are building correctly, but still having some issues with the tests in the Mobile web area.

PR #3211, opened 2018-01-09
CRM Groups schema

andersdd:
Builds on xtuple/xtuple#3205


andersdd:
`groups_type` has been implemented as suggested

PR #3213, opened 2018-01-09
issue #31984 - Revert unnecessary change

ChristopherCotnoir:

PR #3212, opened 2018-01-09
issue #31983 Don't only call function on first row

ChristopherCotnoir:

PR #3214, opened 2018-01-10
fix various beginning-of-year problems

gilmoskowitz:
also added and tested convenience function to dblib to use js metasql parser

this should fix the CI failures on xtuple/xtuple#3212 and xtuple/xtuple#3213


gilmoskowitz:
jenkins - retest this please

PR #3218, opened 2018-01-25
issue #32017 Format shipvia consistently

ChristopherCotnoir:


foundation-database/public/functions/fetchdefaultshipvia.sql
```diff
@@ -4,7 +4,7 @@ CREATE OR REPLACE FUNCTION FetchDefaultShipVia() RETURNS TEXT AS '
 DECLARE
   _returnVal TEXT;
 BEGIN
-  SELECT shipvia_code INTO _returnVal
+  SELECT shipvia_code || '-' || shipvia_descrip INTO _returnVal
```
davidbeauchamp: shipvia_descrip should be COALESCE'd here because it isn't required and could be null.

ChristopherCotnoir: The only place in any of our code base that allows you to insert or update shipvia is in the Ship Vias section of the Sales Setup screen, and that always saves a non-null description. All quickstart/demo data has a non-null description as well. You'd have to do something custom to get a null description, and if you did, you'd have bigger problems, because the COALESCE is already missing for all the Ship Via comboboxes in the entire application, so it would show up blank everywhere already anyway. We can fix this when we change those as well, but for now I figure it is better to be wrong and consistent than right and inconsistent.

davidbeauchamp: I don't follow the rationale here, shipvia_descrip is nullable in the database, and the code that saves the shipvia does not require you enter in a description as you have noticed. As we are all well aware, people INSERT data into the database outside of the interface all the time. I am not on board with leaving it intentionally up to chance whether or not this concatenation works based on those nonexistent controls. If the other comboboxes need to be fixed as well so be it, but I do not think it should be brushed off. Maybe I am alone here.

jwirt:
I've found the functionality of the shipvia combobox to be a problematic interface.  We've written extensions that try to load the shipvia from data we pull from the UPS api.  Some places in the process the shipvia will accept a valid shipvia_code as valid input.  In some other place we have to insert the shipvia_code || - || shipvia_description.  Some versions in the past allowed freeform shipvia. And some REQUIRED shipvia_code || -.  I haven't experienced any other xtuple comboboxes that have had that sort of variations in the type of data it would accept.  I appreciate you guys working on it and I'm trusting that it will function in the same consistent way as the other combo boxes.  My thought would be that if a valid "shipvia_code" was entered then it should be considered valid data.

foundation-database/public/functions/fetchdefaultshipvia.sql
gilmoskowitz: I agree with @davidbeauchamp - at minimum add the COALESCE here.

We still need to address the broader problem of ad hoc ship-vias and resolving the disparities of some places using id, others using concatenated text, yet others using just code, and the occasional case that might be using the description only. This may involve an upgrade script.

foundation-database/public/functions/fetchdefaultshipvia.sql
```diff
@@ -4,7 +4,7 @@ CREATE OR REPLACE FUNCTION FetchDefaultShipVia() RETURNS TEXT AS '
 DECLARE
   _returnVal TEXT;
 BEGIN
-  SELECT shipvia_code INTO _returnVal
+  SELECT shipvia_code || '-' || COALESCE(shipvia_descrip, '') INTO _returnVal
```
gilmoskowitz:
```
psql:/home/ubuntu/jenkins/workspace/xtuple-xtuple-pr/xtuple/scripts/output/build_jenkins14724.sql:27000: ERROR:  syntax error at or near "-"
LINE 7:   SELECT shipvia_code || '-' || COALESCE(shipvia_descrip, ''...
```

ChristopherCotnoir:
jenkins - retest this please

ChristopherCotnoir:
jenkins - retest this please

gilmoskowitz:
See the related bug [25621](https://www.xtuple.org/xtincident/view/bugs/25621)

PR #3236, opened 2018-03-08
issue #31879 Don't block group deletion if the group has privs

ChristopherCotnoir:


ChristopherCotnoir:
Associated with xtuple/qt-client#1878.

PR #3237, opened 2018-03-08
issue #31967 Renamed function so as not to imply it does not modify the database

ChristopherCotnoir:


ChristopherCotnoir:
Associated with xtuple/qt-client#1881.

PR #3243, opened 2018-03-15
Issue #32550 - Move ccbank_ccard_type check to fundstype_code fkey.

bendiy:
Move the hard coded `ccbank_ccbank_ccard_type_check` check constraint `CHECK (ccbank_ccard_type IN ('A', 'D', 'M', 'P', 'V', 'O'))` to a `fundstype_code` foreign key.


foundation-database/public/patches/populate_fundstype.sql
```diff
@@ -187,3 +187,22 @@ SELECT
...
+SELECT
+  'P',
+  'Paypal',
+  'Payment by Paypal',
+  FALSE,
+  FALSE,
+  FALSE
+WHERE NOT EXISTS (
+  SELECT 1 FROM fundstype WHERE fundstype_code = 'P'
+);
```
bendiy: I don't know if we really need to include `P` for PayPal here. It's for a stub in the creditcardprocessor.cpp that hasn't been finished.

PR #3245, opened 2018-03-16
issue #32496, issue #17968 - preserve underlying code value

gilmoskowitz:


gilmoskowitz:
This change is unnecessary and might break reports.

PR #3249, opened 2018-03-21
issue #32135 Fix incorrect query

ChristopherCotnoir:


ChristopherCotnoir:
Associated with xtuple/qt-client#1896.

PR #3253, opened 2018-03-22
 Issue #27306 - credit card preflight checks

bendiy:
Add a ccard.sql file to include set a `ccard_ccard_type_fkey` to `ccbank_ccard_type`.

Add a database function canProcessPaymentGatewayTransaction() that checks if all of the accounting is setup for a customer and payment method source. If not, it raises an excaption explaining what's not seutp.

Add newer company_unassigned_accnt_id column to 440_schema and populate it in demo_data.


foundation-database/public/functions/canProcessPaymentGatewayTransaction.sql
```diff
@@ -0,0 +1,219 @@
...
+  END IF;
+  IF (NOT _custCardAccnt.glaccnt_same_company) THEN
+    RAISE EXCEPTION 'G/L Accounts are not for the same Company. [xtuple: canProcessPaymentGatewayTransaction, -17]';
+  END IF;
+  IF (NOT _custCardAccnt.discrep_accnt_is_valid) THEN
+    RAISE EXCEPTION 'G/L Series Discrepancy Account is not setup. [xtuple: canProcessPaymentGatewayTransaction, -18]';
+  END IF;
+  IF (NOT _custCardAccnt.unassigned_accnt_is_valid) THEN
+    RAISE EXCEPTION 'Unassigned G/L Account is not setup. [xtuple: canProcessPaymentGatewayTransaction, -19]';
+  END IF;
```
gilmoskowitz: The specificity here is good. It would be nice (not required but nice) if only one exception were raised that listed all of the problems encountered. I don't know how to do that and still be able to translate the messages, though. It's worth some thought.

foundation-database/public/functions/canProcessPaymentGatewayTransaction.sql
```diff
@@ -0,0 +1,219 @@
...
+      RAISE EXCEPTION 'Tax Code "%" does not have a G/L Account assigned. [xtuple: canProcessPaymentGatewayTransaction, -20]',
+                      _taxCodeAccnt.tax_code;
+    END IF;
+    IF (NOT _taxCodeAccnt.tax_glaccnt_is_liability) THEN
+      RAISE EXCEPTION 'Tax Code "%" G/L Account "%" is not a Liability account. [xtuple: canProcessPaymentGatewayTransaction, -21]',
+                      _taxCodeAccnt.tax_code, _taxCodeAccnt.accnt_descrip;
+    END IF;
+    IF (NOT _taxCodeAccnt.tax_glaccnt_is_active) THEN
+      RAISE EXCEPTION 'Tax Code "%" G/L Account "%" is not Active. [xtuple: canProcessPaymentGatewayTransaction, -22]',
+                      _taxCodeAccnt.tax_code, _taxCodeAccnt.accnt_descrip;
```
gilmoskowitz: Here and above -- these calls to `raise exception` don't pass their parameters in a way that can be used by a translation. To do so, write them this way:
```
RAISE EXCEPTION 'Tax Code "%" G/L Account "%" is not Active. [xtuple: canProcessPaymentGatewayTransaction, -22, %, %]',
                _taxCodeAccnt.tax_code, _taxCodeAccnt.accnt_descrip,
                _taxCodeAccnt.tax_code, _taxCodeAccnt.accnt_descrip;
```

foundation-database/public/functions/canProcessPaymentGatewayTransaction.sql
```diff
@@ -0,0 +1,219 @@
...
+  pCurrId integer
+) RETURNS boolean AS
+$BODY$
+-- Copyright (c) 1999-2018 by OpenMFG LLC, d/b/a xTuple.
+-- See www.xtuple.com/CPAL for the full text of the software license.
+DECLARE
+  _currentAccntPeriod RECORD;
+  _custCardAccnt RECORD;
+  _exchangeRate NUMERIC;
+  _isMmltiCurrency BOOLEAN := false;
```
gilmoskowitz: typo - `m` should be `u`, but see comment below - this variable isn't necessary.

foundation-database/public/functions/canProcessPaymentGatewayTransaction.sql
```diff
@@ -0,0 +1,219 @@
...
+
+  IF (_currentAccntPeriod.period_freeze) THEN
+    RAISE EXCEPTION 'Cannot post to a frozen Accounting Period for "%". [xtuple: canProcessPaymentGatewayTransaction, -25]',
+                    CURRENT_DATE;
+  END IF;
+
+  IF (_isMmltiCurrency) THEN
+    -- Will throw an error if the exchange rate is not setup.
+    _exchangeRate := currRate(pCurrId, baseCurrID(), CURRENT_DATE);
+  END IF;
```
gilmoskowitz: There should be an exchange rate set to 1 from `startOfTime()` to `endOfTime()` for the base currency. Therefore, the multicurrency check above and the conditional here aren't necessary. Simply
```
PERFORM currRate(pCurrId, baseCurrId(), CURRENT_DATE);
```

foundation-database/public/functions/canProcessPaymentGatewayTransaction.sql
```diff
@@ -0,0 +1,219 @@
...
+
+  IF (_custCardAccnt IS NULL) THEN
+    RAISE EXCEPTION 'Invaid Customer or Payment Method. [xtuple: canProcessPaymentGatewayTransaction, -1]';
+  END IF;
+  IF (NOT _custCardAccnt.card_has_ccbank) THEN
+    RAISE EXCEPTION 'Payment Method Bank Account is not mapped for ccard_type "%". [xtuple: canProcessPaymentGatewayTransaction, -2]',
+                    _custCardAccnt.ccard_type;
+  END IF;
+  IF (NOT _custCardAccnt.ccbank_has_bankaccnt) THEN
+    RAISE EXCEPTION 'Payment Method Bank Account is not setup for ccard_type "%". [xtuple: canProcessPaymentGatewayTransaction, -3]',
```
gilmoskowitz: typo - `setup` should be `set up` (noun vs. verb)

foundation-database/public/functions/canProcessPaymentGatewayTransaction.sql
```diff
@@ -0,0 +1,219 @@
...
+  IF (NOT _custCardAccnt.bankaccnt_glaccnt_is_asset) THEN
+    RAISE EXCEPTION 'Payment Method Bank Account "%" G/L Account "%" is not an Asset account. [xtuple: canProcessPaymentGatewayTransaction, -6]',
+                    _custCardAccnt.bankaccnt_name, _custCardAccnt.bankaccnt_accnt_descrip;
+  END IF;
+  IF (NOT _custCardAccnt.bankaccnt_glaccnt_is_active) THEN
+    RAISE EXCEPTION 'Payment Method Bank Account "%" G/L Account "%" is not Active. [xtuple: canProcessPaymentGatewayTransaction, -7]',
+                    _custCardAccnt.bankaccnt_name, _custCardAccnt.bankaccnt_accnt_descrip;
+  END IF;
+  IF (NOT _custCardAccnt.cust_has_ar_glaccnt) THEN
+    RAISE EXCEPTION 'Customer "%" does not have a A/R G/L Account assigned. [xtuple: canProcessPaymentGatewayTransaction, -8]',
```
gilmoskowitz: typo - `a A/R` should be `an A/R`

foundation-database/public/functions/canProcessPaymentGatewayTransaction.sql
```diff
@@ -0,0 +1,219 @@
...
+    RAISE EXCEPTION 'Customer "%" Deferred Revenue Account "%" is not a Liability account. [xtuple: canProcessPaymentGatewayTransaction, -15]',
+                    _custCardAccnt.cust_number, _custCardAccnt.deferred_accnt_descrip;
+  END IF;
+  IF (NOT _custCardAccnt.deferred_glaccnt_is_active) THEN
+    RAISE EXCEPTION 'Customer "%" Deferred Revenue Account "%" is not Active. [xtuple: canProcessPaymentGatewayTransaction, -16]',
+                    _custCardAccnt.cust_number, _custCardAccnt.deferred_accnt_descrip;
+  END IF;
+  IF (NOT _custCardAccnt.glaccnt_same_company) THEN
+    RAISE EXCEPTION 'G/L Accounts are not for the same Company. [xtuple: canProcessPaymentGatewayTransaction, -17]';
+  END IF;
```
gilmoskowitz: - Is there an explicit check somewhere in the CC processing for the company segments of the two G/L accounts matching? If so then this check is fine.
- It would be helpful for the message to say which accounts don't match, as `accnt` is joined with 6 times.

scripts/lib/build_all.js
```diff
@@ -8,7 +8,9 @@ var _ = require('underscore'),
   buildDictionary = require("./build_dictionary"),
   buildClient = require("./build_client").buildClient,
   defaultExtensions = require("./util/default_extensions").extensions,
-  initDatabase = require("./util/init_database").initDatabase,
+  init_database = require("./util/init_database"),
```
gilmoskowitz: We're trying to get rid of build_app, not enhance its capabilities. The one small update could be much more concisely handled with less risk of breakage by modifying demo_data.sql. That would eliminate 4 files and 51 lines worth of change here plus https://github.com/xtuple/private-extensions/pull/999

bendiy: `demo_data.sql` is applied to a 440 schema, which does not have the column I need to populate.
I can modify the 440 schema by adding the missing column in `demo_data.sql` if you would prefer.

gilmoskowitz: Please add the column at the end of 440_schema with a comment about why it's here. That'll keep the reference database builds in sync. We're trying to get rid of `440_schema.sql` and build the quickstart db from the empty db and demo from quickstart, so this will be short-lived and still a much smaller set of changes.

foundation-database/public/functions/canProcessPaymentGatewayTransaction.sql
bendiy: The `for the same Company` check is to avoid getting this error:
https://github.com/xtuple/xtuple/blob/4_11_x/foundation-database/public/functions/postglseries.sql#L49-L70

foundation-database/public/functions/canProcessPaymentGatewayTransaction.sql
bendiy: This could return an array or map of the errors, but the `storedProcErrorLookup` or `ErrorReporter::getQueryStoredProcErrorMsg()` would then need to understand how to deconstruct that list.

Another possible solution would be to return one error message that contains all the checks with some kind of `OK` and `FAILURE` prefix. e.g.

```
...
OK - Customer "%" does not have a Prepaid G/L Account assigned. [xtuple: canProcessPaymentGatewayTransaction, -11]
OK - Customer "%" Prepaid Account "%" is not a Revenue account. [xtuple: canProcessPaymentGatewayTransaction, -12]
FAILURE - Customer "%" Prepaid Account "%" is not Active. [xtuple: canProcessPaymentGatewayTransaction, -13]
OK - Customer "%" does not have a Deferred Revenue G/L Account assigned. [xtuple: canProcessPaymentGatewayTransaction, -14]
...
```

That could end up being a long message that doesn't fit on a narrow screen, so you can't click 'OK'.

If I refactor this into smaller functions that can be called elsewhere, those single errors will be triggered instead of a list.

I initially considered throwing only one error. In most of these cases, I expect users will only encounter these error the first few times they use the system. Then once configured correctly, only see the occasional `Customer "%" does not have a A/R G/L Account assigned.` message for new customers that were not completely setup.

foundation-database/public/functions/canProcessPaymentGatewayTransaction.sql
```diff
@@ -0,0 +1,219 @@
...
+                    _custCardAccnt.bankaccnt_name;
+  END IF;
+  IF (NOT _custCardAccnt.bankaccnt_has_glaccnt) THEN
+    RAISE EXCEPTION 'Payment Method Bank Account "%" does not have a G/L Account assigned. [xtuple: canProcessPaymentGatewayTransaction, -5]',
+                    _custCardAccnt.bankaccnt_name;
+  END IF;
+  IF (NOT _custCardAccnt.bankaccnt_glaccnt_is_asset) THEN
+    RAISE EXCEPTION 'Payment Method Bank Account "%" G/L Account "%" is not an Asset account. [xtuple: canProcessPaymentGatewayTransaction, -6]',
+                    _custCardAccnt.bankaccnt_name, _custCardAccnt.bankaccnt_accnt_descrip;
+  END IF;
```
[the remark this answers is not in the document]
bendiy: The list I'm working off of is the list @ChristopherCotnoir created on the incident. The account `active` and `type` checks are added to satisfy the `Make sure XYZ is a valid account`. For the account `type`, I the desktop client UI enforces that when setting most of these accounts.

https://github.com/xtuple/qt-client/blob/4_11_x/guiclient/arAccountAssignment.cpp#L39-L43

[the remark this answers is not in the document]
gilmoskowitz: The governing principle needs to be, "Will skipping this check now allow storing the credit card transaction to fail?" If the answer is "No, the credit card transaction will still be saved to the database without errors," then the check does not belong here.

foundation-database/public/functions/canProcessPaymentGatewayTransaction.sql
```diff
@@ -0,0 +1,219 @@
...
+                    _custCardAccnt.ccard_type;
+  END IF;
+  IF (NOT _custCardAccnt.ccbank_has_bankaccnt) THEN
+    RAISE EXCEPTION 'Payment Method Bank Account is not setup for ccard_type "%". [xtuple: canProcessPaymentGatewayTransaction, -3]',
+                    _custCardAccnt.ccard_type;
+  END IF;
+  IF (NOT _custCardAccnt.bankaccnt_is_revenue) THEN
+    RAISE EXCEPTION 'Payment Method Bank Account "%" is not used for accounts receivable. [xtuple: canProcessPaymentGatewayTransaction, -4]',
+                    _custCardAccnt.bankaccnt_name;
+  END IF;
```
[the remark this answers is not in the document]
bendiy: This is a desktop client UI enforced restriction by using `XComboBox::ARBankAccounts`:
https://github.com/xtuple/qt-client/blob/4_11_x/guiclient/configureCC.ui#L293

The check is for `Make sure the default bank account associated with the credit card type is a valid account`.

I can remove these non-stored procedure checks if you would like. The chance that users run into them is low since the UI prevents the failure mode from being selected during setup.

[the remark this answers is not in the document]
bendiy: Saw your latest comment. I'll remove them.

foundation-database/public/functions/canProcessPaymentGatewayTransaction.sql
```diff
@@ -0,0 +1,184 @@
...
+    bankaccnt_name,
+    ccbank_ccard_type IS NOT NULL AS card_has_ccbank,
+    bankaccnt_id IS NOT NULL AS ccbank_has_bankaccnt,
+    bankaccnt_accnt_id IS NOT NULL AS bankaccnt_has_glaccnt,
+    accntIsValid(bankaccnt_accnt_id) AS bankaccnt_glaccnt_is_valid,
+    findARAccount(cust_id) > 0 AS cust_has_ar_glaccnt,
+    accntIsValid(findARAccount(cust_id)) AS ar_glaccnt_is_valid,
+    findPrepaidAccount(cust_id) > 0 AS cust_has_prepaid_glaccnt,
+    accntIsValid(findPrepaidAccount(cust_id)) AS prepaid_glaccnt_is_valid,
+    dfindDeferredAccount(cust_id) > 0 AS cust_has_deferred_glaccnt,
```
gilmoskowitz: There's a bug here - extra `d` - that's causing CI to fail for xtuple/payment-gateways#23.

bendiy: Merged a fix:
https://github.com/xtuple/xtuple/pull/3256
Rerunning CI on https://github.com/xtuple/payment-gateways/pull/23

PR #3255, opened 2018-03-23
issue #32629 Give locks the same state_change as the parent so that nesting happen…

ChristopherCotnoir:
…s properly


foundation-database/public/tables/metasql/processes-detail.mql
```diff
@@ -31,7 +31,8 @@ SELECT CASE WHEN lock_id IS NOT NULL THEN lock_id ELSE pid END AS id, type,
                                   THEN <? value("disabled") ?>
                                   END
                                   ELSE NULL END AS state,
-       CASE WHEN type=0 THEN state_change ELSE NULL END AS state_change,
+       state_change,
+       CASE WHEN type=0 THEN state_change ELSE NULL END AS state_change_qtdisplayrole,
```
gilmoskowitz: This will [always show the `state_change`](https://github.com/xtuple/qt-client/blob/master/widgets/xtreewidget.cpp#L549).

PR #3260, opened 2018-03-28
issue #32675 - fix syntax err when specifying primary key

gilmoskowitz:


davidbeauchamp:
Not to confuse things even further but there is this function too [xt.add_primary_key](https://github.com/xtuple/xtuple/blob/4_11_x/lib/orm/source/xt/functions/add_primary_key.sql)

PR #3265, opened 2018-04-06
issue #32648 Don't silently wipe shipto on sales order either

ChristopherCotnoir:

PR #3267, opened 2018-04-09
Change to outer join

ChristopherCotnoir:


ChristopherCotnoir:
Change made in response to comment on similar issue in xtuple/xtuple#3265.

PR #3274, opened 2018-04-20
issue #30503 Add contact marking

ChristopherCotnoir:


ChristopherCotnoir:
Associated with xtuple/qt-client#1921.

PR #3276, opened 2018-04-26
issue #32903 Use checkitem_curr_rate to calculate currency gain/loss

ChristopherCotnoir:


harithamekala:
Performed UAT: Observed that when user tries to void an AP Check the system doesn’t returns any error. Followed the below steps 
 
Steps:
1.	Open ‘Payables Workbench’ screen from ‘Accounting->AccountsPayable->Workbench’.
2.	Clicks on ‘New Misc.’ button.
3.	Ensure that ‘Miscellaneous Voucher’ screen is displayed.
4.	Enter Vendor say ‘TPARTS’, ‘Amount to Distribute’ say ‘100’ in ‘EUR’
5.	Navigate to ‘Misc.Distributions’ tab. Clicks on ‘New’ button 
6.	Enter ‘Account’ and clicks on ‘Save’ button, also enters ‘Invoice Date’ and tab out of the field.
7.	Clicks on ‘Save’ and ‘Cancel’ buttons.
8.	Navigate to ‘Payables’ tab and clicks on ‘Query’ 
9.	Selects the voucher and clicks on ‘Approve’ button.
10.	Ensure that ‘Approve payment’ screen is opened.
11.	Selects ‘Bank Account’ says ‘EBANK- eBank Checking Account’ from the dropdown and clicks on ‘Save’ button.
12.	‘Currencies Do Not Match’ pop up is displayed. Clicks on ‘Yes’ button.
13.	Navigate to ‘Payments’ tab, clicks on ‘Prepare’ button.
14.	‘Prepare Payment’ screen is displayed, clicks on ‘Prepare’ button.
15.	Selects the Payment and clicks on ‘Void’ button.
16.	Observed that voucher is voided successfully. Please refer to the attached screenshot ‘voucherVoid.png’

vouchervoid

foundation-database/public/patches/fixcheckitemcurrrate.sql
```diff
@@ -1,11 +1,139 @@
...
+                       GROUP BY checkitem_curr_id),
+             rate AS (SELECT checkitem_curr_rate * _r.checkhead_curr_rate AS actual, stored
+                        FROM checkitem
+                        JOIN curr ON checkitem.checkitem_curr_id = curr.checkitem_curr_id
+                       WHERE checkitem_checkhead_id = _r.checkhead_id)
+        UPDATE checkitem
+           SET checkitem_curr_rate = CASE WHEN ROUND(rate.actual, 6) = ROUND(rate.stored, 6)
+                                          THEN rate.stored
+                                          ELSE rate.actual
+                                      END
```
gilmoskowitz: Why bother with `rate.stored`? The `CASE` says to use `rate.actual` unless they're approximately the same anyway.

ChristopherCotnoir: rate.actual is the result of the multiplication of two values (one of which was historically derived through division). If it is close enough to rate.stored that implies that nobody was messing about with the currency exchange rates and the difference is due to rounding, so it would be best to use the stored rate. We only want to override that if there is reason to believe they are genuinely different. It's an accuracy issue. I was unsure how many decimal places to check here, but 6 is what I went with.

foundation-database/public/patches/fixcheckitemcurrrate.sql
```diff
@@ -1,11 +1,139 @@
...
+              FROM checkhead
+             WHERE EXISTS (SELECT 1
+                             FROM checkitem
+                            WHERE checkitem_checkhead_id = checkhead_id
+                              AND checkitem_curr_id != _base)
+  LOOP
+    SELECT MAX(checkitem_curr_id) AS curr, COUNT(DISTINCT checkitem_curr_id) AS count,
+           COUNT(checkitem_id) AS countitems INTO _curr
+      FROM checkitem
+     WHERE checkitem_checkhead_id = _r.checkhead_id;
```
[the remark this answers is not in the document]
ChristopherCotnoir: In most databases the number of entries in the loop will be very small (or even 0) compared to the total number of checkhead entries. I don't want to query this information except when it is actually needed.

PR #3290, opened 2018-05-15
[32307] Prevent spurious inventory history records for item cost changes

andersdd:
where no value change is posted.


ChristopherCotnoir:
jenkins - retest this please

PR #3292, opened 2018-05-17
[32927] Invoice line item ordering for kits

andersdd:


srigurudatta:
Performed UAT. Issue is fixed. When printing an invoice with kit items, the kit part numbers are displaying first and followed by the kit contents. Please find the attached files ‘Invoice1.png’ and ‘Invoice2.png
invoice1
’.
invoice2

PR #3299, opened 2018-05-24
[32560] Always match .*

ChristopherCotnoir:


foundation-database/public/functions/itemcharprice.sql
```diff
@@ -179,13 +179,13 @@ BEGIN
           -- 2. Specific Customer Shipto Pattern
           OR (COALESCE(LENGTH(ipsass_shipto_pattern), 0) > 0
               AND ipsass_cust_id > -1
-              AND ipsass_shipto_pattern ~ _shipto.shipto_num
+              AND ipsass_shipto_pattern ~ COALESCE(_shipto.shipto_num, '')
```
gilmoskowitz: This should never have worked. `~` isn't commutative:
```
# select '.*' ~ '0', '0' ~ '.*';
 ?column? | ?column? 
----------+----------
 f        | t
(1 row)
```

ChristopherCotnoir: It WAS working, when I first looked at this code, but Ben's recent refactor PR switched the order for some reason and I didn't notice when applying this change. Good catch.

ChristopherCotnoir:
jenkins - retest this please

PR #3301, opened 2018-05-25
[33130] Properly copy item groups, characteristic default and price, and bomi…

ChristopherCotnoir:
…tem_issuewo


ChristopherCotnoir:
Associated with xtuple/xdruple-extension#117.

PR #3313, opened 2018-06-11
[33238] Prevent ability to receipt into inactive item sites.

andersdd:
Also prevents amending an item type if the item site validation checks would fail


foundation-database/public/functions/enterreceipt.sql
```diff
@@ -45,6 +45,7 @@ BEGIN
 	   CASE WHEN (poitem_itemsite_id = -1) THEN NULL
 		ELSE poitem_itemsite_id
 	   END AS itemsite_id,
+           COALESCE(itemsite_active, true) AS itemsite_active,
```
gilmoskowitz: This allows P/O receipts into non-existent itemsites.

PR #3314, opened 2018-06-12
[33194] Copy Item user costs when copying items

andersdd:
User costs were missed when copying items.

### DIFF
--- a/foundation-database/public/functions/copyitem.sql
+++ b/foundation-database/public/functions/copyitem.sql
@@ -1,5 +1,5 @@
 CREATE OR REPLACE FUNCTION copyItem(INTEGER, TEXT) RETURNS INTEGER AS $$
--- Copyright (c) 1999-2016 by OpenMFG LLC, d/b/a xTuple.
+-- Copyright (c) 1999-2018 by OpenMFG LLC, d/b/a xTuple.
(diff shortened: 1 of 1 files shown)


lalithapadala:
Performed UAT: Issue still exists. User defined costs are not copied to the target item. Please refer to the attached screenshot.

ChristopherCotnoir:
Also be wary of this:

https://github.com/xtuple/private-extensions/blob/4_12_x/source/manufacturing/foundation-database/xtmfg/functions/copyitem.sql

There may be search path issues to deal with because xtte also overloads copyitem. This came up when trying to overload copyItem for xdruple.

andersdd:
Thank you for the comments @gilmoskowitz and @ChristopherCotnoir .  Not entirely sure what is going on in the test situations here.  I am developing/building on a manufacturing system and did not encounter issues as described.  Also note @lalithapadala screen shots copying _some_ user defined cost elements, but not all which is not expected.  I need to invesitgate further...

andersdd:
Rechecked code and all overlay functions in scripted packages call the public function first.  The `pCopyCosts` is basically ignored as the costs are copied regardless.

Not sure about the UAT failure as it works nicely on my system and the code would suggest it always should work. (Cannot retest until the scripted setup bug is resolved).

PR #3328, opened 2018-06-29
issue #32336 - added checks to invoiceline.sql and created test for invoiceline view

mmicr001:
- Created `test/database/views/invoiceline.js` to test invoiceline view
- Added checks to see if record was found on insert and update functions.

### DIFF
--- a/test/database/views/invoiceline.js
+++ b/test/database/views/invoiceline.js
@@ -1,0 +1,143 @@
+var _ = require("underscore"),
+  assert = require('chai').assert,
+  path = require('path');
+
+(function () {
+  "use strict";
+  describe('api.invoiceline view', function () {
+
+    var loginData = require(path.join(__dirname, "../../lib/login_data.js")).data,
+      datasource = require('../../../node-datasource/lib/ext/datasource').dataSource,
+      config = require(path.join(__dirname, "../../../node-datasource/config.js")),
+      creds = _.extend({}, config.databaseServer, {database: loginData.org});
+
+    function invoiceline(obj) {
+      return "ROW('" + obj.invoice_number        + "',"
+                     + (obj.line_number           ?       obj.line_number           + ","  : "1,")
+                     + (obj.item_number           ? "'" + obj.item_number           + "'," : "NULL,")
+                     + (obj.misc_item_number      ? "'" + obj.misc_item_number      + "'," : "NULL,")
+                     + (obj.site                  ? "'" + obj.site                  + "'," : "'WH1',")
+                     + (obj.misc_item_description ? "'" + obj.misc_item_description + "'," : "NULL,")
+                     + (obj.sales_category        ?       obj.sales_category        + "'," : "NULL,")
+                     + (obj.customer_part_number  ?       obj.customer_part_number  + "'," : "NULL,")
+                     + (obj.qty_ordered           ?       obj.qty_ordered           + ","  : "0,")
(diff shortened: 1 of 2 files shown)


test/database/views/invoiceline.js
```diff
@@ -0,0 +1,143 @@
...
+    });
+
+    it("should reject insert for non-existent invoice", function (done) {
+      var line = invoiceline({invoice_number: 'non-existent'}),
+          sql  = "select insertInvoiceLineItem(%) as result;".replace(/%/g, line)
+        ;
+      datasource.query(sql, creds, function (err, res) {
+        console.error(err);
+        assert.isNotNull(err);
+        assert.match(err, /xtuple:.*-1[^0-9]|not found/i);
```
gilmoskowitz: `dblib` provides a function for this - [assertErrorCode](https://github.com/xtuple/xtuple/blob/4_12_x/test/database/dblib.js#L19). If you're feeling brave, it would be nice to add this to the `assert` prototype, say as `xtupleDbErr`. That would allow us to write this check three ways:
```JavaScript
assert.isNotNull(err); assert.match(err, /xtuple:.*-1[^0-9]*|not found/i); -- 1
dblib.assertErrorCode(err, res, 'insertInvoiceLineItem', -1);              -- 2
assert.xtupleDbErr(err, res, 'insertInvoiceLineItem', -1);                 -- 3
```

test/database/views/invoiceline.js
```diff
@@ -0,0 +1,143 @@
...
+      var invoice = "ROW('99999', NULL, NULL, NULL, NULL,"
+               +      "NULL, NULL, 0,"
+               +      "NULL, NULL, 'TTOYS',"
+               +      "NULL, NULL, NULL, NULL, NULL, "
+               +      "NULL, NULL, NULL, NULL, NULL, "
+               +      "NULL, NULL, NULL, NULL, NULL, "
+               +      "NULL, NULL, NULL, NULL, NULL, "
+               +      "NULL, NULL, NULL, NULL, NULL, "
+               +      "0, NULL, 0, 'USD' , 0, 'invoice notes')",
+          sql  = "select insertInvoice(%) as result;".replace(/%/g, invoice)
```
gilmoskowitz: This works. See the parameter binding approach used in some [other places](https://github.com/xtuple/xtuple/blob/4_12_x/test/database/functions/addrusecount.js#L35-L38), too. It's generally safer to use parameter binding, so that's what we should do in production code. In these tests it's not as important and we can use whichever is more convenient.

test/database/views/invoiceline.js
```diff
@@ -0,0 +1,143 @@
...
+        assert.isNotNull(err);
+        assert.match(err, /xtuple:.*-1[^0-9]|not found/i);
+        done();
+      });
+    });
+
+   
+
+     it("should allow creating a new invoice", function (done) {
+      var invoice = "ROW('99999', NULL, NULL, NULL, NULL,"
```
gilmoskowitz: To reduce the risk of conflicts and unanticipated errors, it's generally safer to `fetchInvcNumber` and save it in a JS variable than to assume
- 99999 is unused
- the test will clean up properly after a failure

See [some of](https://github.com/xtuple/xtuple/blob/48f9ecb8733685e2754c5a145fc8ac4c96b12506/test/database/functions/voidcreditmemo.js#L37) the [other tests](https://github.com/xtuple/xtuple/blob/4_12_x/test/database/functions/adjustinvvalue.js#L16-L34) for examples of these "it.needs something" calls.

PR #3337, opened 2018-07-11
[32453] Add option to suppress automatic supply order creation, fix issue whe…

ChristopherCotnoir:
…re only first supply order is created from places other than sales order item screen

### DIFF
--- a/foundation-database/public/tables/itemsite.sql
+++ b/foundation-database/public/tables/itemsite.sql
@@ -62,7 +62,8 @@ SELECT
   xt.add_column('itemsite', 'itemsite_recvlocation_dist',  'BOOLEAN', 'DEFAULT false NOT NULL', 'public'),
   xt.add_column('itemsite', 'itemsite_issuelocation_dist', 'BOOLEAN', 'DEFAULT false NOT NULL', 'public'),
   xt.add_column('itemsite', 'itemsite_created',      'TIMESTAMP WITH TIME ZONE', NULL, 'public'),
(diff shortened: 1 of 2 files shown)


ChristopherCotnoir:
Requires xtuple/qt-client#1987.

foundation-database/public/tables/itemsite.sql
```diff
@@ -62,7 +62,8 @@ SELECT
   xt.add_column('itemsite', 'itemsite_recvlocation_dist',  'BOOLEAN', 'DEFAULT false NOT NULL', 'public'),
   xt.add_column('itemsite', 'itemsite_issuelocation_dist', 'BOOLEAN', 'DEFAULT false NOT NULL', 'public'),
   xt.add_column('itemsite', 'itemsite_created',      'TIMESTAMP WITH TIME ZONE', NULL, 'public'),
-  xt.add_column('itemsite', 'itemsite_lastupdated',  'TIMESTAMP WITH TIME ZONE', NULL, 'public');
+  xt.add_column('itemsite', 'itemsite_lastupdated',  'TIMESTAMP WITH TIME ZONE', NULL, 'public'),
+  xt.add_column('itemsite', 'itemsite_noautoord',          'BOOLEAN', 'NOT NULL DEFAULT FALSE', 'public');
```
gilmoskowitz: It's _much_ safer to name boolean variables as positives:
```
xt.add_column('itemsite', 'itemsite_autoord',          'BOOLEAN', 'NOT NULL DEFAULT TRUE', 'public');
```
This avoids 2 double negatives in the coitem trigger below and one here, with the cost of 3 lines to set the new checkbox's default value in xtuple/qt-client#1987

PR #3345, opened 2018-07-13
fix 'mrkd_id must be in group-by' error

gilmoskowitz:
and incidentally problematic references to contact created and updated columns

### DIFF
--- a/foundation-database/public/tables/metasql/contacts-detail.mql
+++ b/foundation-database/public/tables/metasql/contacts-detail.mql
@@ -1,14 +1,15 @@
 -- Group: contacts
 -- Name: detail
 -- Notes: 
--- Copyright (c) 1999-2018 by OpenMFG LLC, d/b/a xTuple.
--- See www.xtuple.com/CPAL for the full text of the software license.
+--        Copyright (c) 1999-2018 by OpenMFG LLC, d/b/a xTuple.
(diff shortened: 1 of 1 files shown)


foundation-database/public/tables/metasql/contacts-detail.mql
```diff
@@ -37,7 +38,7 @@ SELECT DISTINCT ON (length(cntct_last_name)=0, length(cntct_first_name)=0, cntct
 <? endforeach ?>
 <? endif ?>
 <? if exists("hasContext") ?>
-FROM cntct
+  FROM cntct
```
andersdd: Should this really be cntct() to support CRM privileges??

gilmoskowitz: It appears from code history that `hasContext` is supposed to do this:
```
commit 93b206a1e242c11b2f1941092c41924b0686f570
Author: jrogelstad <jrogelstad>
Date:   Mon Sep 12 22:58:29 2011 +0000

    issue# 15635 and issue#15501: Add context option for when to-do lists exist in CRM or Customer account windows. Show everything in the context, but don't allow editing unless privileges have been granted.
    
    git-svn-id: http://svn.code.sf.net/p/postbooks/code/xtuple/trunk@12431 ea1a15ef-c238-0410-93cb-fa28e9f07733
```

PR #3362, opened 2018-07-27
[33100] Create WO checkbox in Sales Order Line was not working.  

andersdd:
Reinstate logic to create WOs.  Added check to prevent duplicate SO->WOs being created but this hard check needs to be validated.


foundation-database/public/trigger_functions/coitem.sql
```diff
@@ -369,6 +343,36 @@ BEGIN
...
+                    NEW.coitem_id,
+		    cohead_prj_id) INTO NEW.coitem_order_id
+    FROM cohead, itemsite, custinfo
+    WHERE cohead_id=NEW.coitem_cohead_id
+      AND itemsite_id=NEW.coitem_itemsite_id
+      AND cust_id=cohead_cust_id
+      AND NOT EXISTS (SELECT 1 FROM wo
+                       WHERE wo_number  = CAST(cohead_number AS INTEGER)
+                         AND wo_ordtype = 'S'
+                         AND wo_ordid   = NEW.coitem_id);
```
ChristopherCotnoir: This isn't necessary. Once a supply order is created, coitem_order_id is set to that supply order id which prevents this code from running again since coitem_order_id is no longer -1.

ChristopherCotnoir:
The code for PO and PR creation should be moved outside the check for TG_OP = 'INSERT' as well.

foundation-database/public/trigger_functions/coitem.sql
andersdd: Possibly but we have another issue reported about duplicated WOs so I put in a solid check to prevent

andersdd:
PO and PR creation is independent of TG_OP now

PR #3363, opened 2018-07-30
[24325] Prevent zero (or less) Vendor UOM ratio in Item Source record

andersdd:


lalithapadala:
Performed UAT: Issue is fixed. ‘Cannot Save Item Source’ dialog is displayed on saving ‘Item Source’ with ‘Inventory/Vendor UOM Ratio’ value less than or equals to ‘0’. Please refer to the attached screenshot.
uomratio

PR #3390, opened 2018-08-10
[33646] Resolve issue where Ship Shipment was not posting to the GL

andersdd:
This is not a pretty solution as it results in an additional Db call, but the dependencies in the flow of data due to Sales Reservations require shipitems to be created before reservations are processed, before we obtain the transaction value from inventory history.  Only at that point do we know what the shipitem value should be.

realted to xtuple/xtuple#3288


lalithapadala:
Performed UAT: Issue is fixed. Sales Order shipment transaction is posted to GL. Please refer to the attached screenshot.
gltransaction

PR #3399, opened 2018-08-15
Upgrade checkPrivilege() function to handle multiple inputs

andersdd:
Made this change to make resolution of another issue a bit easier.


andersdd:
comment addressed in PR xtuple/xtuple#3405

PR #3402, opened 2018-08-16
[26036] Employee images hitting foreign key issue on inherited tables

andersdd:
Had to remove the FK constraint as it does not work with images source from child package tables


harithamekala:
Performed UAT: Issue is fixed.User is able to select the Employee 'Image' without any crash.

gilmoskowitz:
jenkins - retest this please

PR #3406, opened 2018-08-22
[33723] Change schema introspection query to parameterize the ORM column strings.

bendiy:
The Discovery Doc query is failing after installing xdruple-extension because of this change:
https://github.com/xtuple/xdruple-extension/commit/c28730463beb2c7d2f22686ec61982a8d7cf5897

This parameterization fixes that.


bendiy:
This PR in included in https://github.com/xtuple/xtuple/pull/3407. CI is passing with this PR's change now that Zombie is out of the way.

PR #3409, opened 2018-08-29
Upgrade to Node 8.x

bendiy:
Upgrade to Node 8.x and update dependencies to work with it. Fix tests.


bendiy:
CI will not run/pass until we move the Jenkins servers to Node.js v8.11.4, which is the current LTS release.
I'd recommend installing [nvm](https://github.com/creationix/nvm#installation) on the server to make it easy to switch between Node.js versions.

There is a [Windows version of nvm](https://github.com/coreybutler/nvm-windows) I have had success with, which keeps the same commands/api as the linux/mac version. I have found both of these `nvm` managers to be easy to work with compared to the old [n](https://github.com/tj/n) we used to use.

gilmoskowitz:
jenkins - retest this please

PR #3422, opened 2018-09-20
[33899] Fully define column data type to timestamp with time zone

bendiy:
After adding the `xdruple.oauth_2_token` and `xdruple.oauth_2_client` views, when `add_column()` runs on the `xt.oa2token` and `xt.oa2client` table, we get this error:

```sql
ERROR:  cannot alter type of a column used by a view or rule
DETAIL:  rule _RETURN on view oauth_2_client depends on column "oa2client_issued"
CONTEXT:  SQL statement "ALTER TABLE xt.oa2token ALTER COLUMN oa2token_auth_code_expires SET DATA TYPE timestamp"
```

This appears to be happening because `timestamp` is converted to `timestamp without time zone` and `add_column()`'s check for column type differences sees this as a change.

This PR fully defines the column types as `timestamp with time zone` and drops the dependent views before running `add_column()`. Those views will then be added back later in the build process. On the next build, `add_column()` will not detect a type change for these columns.

### DIFF
--- a/database/source/xt/tables/oa2client.sql
+++ b/database/source/xt/tables/oa2client.sql
@@ -1,61 +1,27 @@
--- table definition
+DROP VIEW IF EXISTS xdruple.oauth_2_token;
+DROP VIEW IF EXISTS xdruple.oauth_2_client;
 
-select xt.create_table('oa2client');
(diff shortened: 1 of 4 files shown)


bendiy:
jenkins - retest this please

PR #3433, opened 2018-09-28
[25092] Fix urlinfo fkey migration to file_id sequence

bendiy:
Fixes CI test rebuild from `xdruple.urlinfo_ext` fkey.


andersdd:
@gilmoskowitz `SET CONSTRAINTS ALL DEFERRED` _does_ work but it appears only on foreign keys with a FK set to `DEFERRABLE`.  The referenced PR makes the offending FK deferrable and my testing.  It would seem only on package presently references urlinfo, so we should not put too much effort into a change that will occur only during one upgrade

PR #3438, opened 2018-10-02
[33936] Item Costs by Class Code - fix discrepancy between screen and report

andersdd:
Added same tolerance found in screen metaSQL to the report SQL


lalithapadala:
Unable to reproduce the issue in 4.11.3 & 4.12.0 QA databases to test the pull request. Please refer to attached files.
itemsbyclasscodereport
printreport

andersdd:
@lalithapadala Create a new item and create an item cost for a cost element of 0.0001.  Do not post to Standard.  Previously that item would be visible in the report but not the screen.  With this change, it should not be visible on either screen or report as the cost difference does not exceed the reporting tolerance.

PR #3449, opened 2018-10-12
[34009] Turn KeyGen script into a console application with a DB connection

amikheychik:
Script can be run same way as `build_app.js`:

```
DATABASE=pilot
CONFIG_DIR=/etc/xtuple/${DATABASE}
APPLICATION=pilot.example.com
ISSUER=pilot
KEY=pilot
./scripts/keygen.js \
        -c ${CONFIG_DIR}/config.js \
        -d ${DATABASE} \
        -a ${APPLICATION} \
        -i ${ISSUER} \
        -p /var/xtuple/keys/${KEY}.p12
```

This can be used to replace https://github.com/xtuple/xtuple-admin-utility/blob/master/functions/oatoken.fun and related calls.

### DIFF
--- a/scripts/keygen.js
+++ b/scripts/keygen.js
@@ -104,28 +104,22 @@
   };
 
   let erp = new DataSource();
-  Promise.resolve()
-    .then(() => {
-      return new Promise(Promises.createKeypair);
-    })
-    .then((keypair) => {
-      Promise.resolve()
-        .then(() => {
-          return new Promise(Promises.readOAuth2ClientSQLFile);
-        })
-        .then((sql) => {
-          return erp.execute(new Query(sql, [
-            script.iss,
-            script.iss,
-            forge.pki.publicKeyToPem(keypair.publicKey)
-          ]));
-        })
-        .then(() => {
-          return new Promise(Promises.saveP12File(keypair.privateKey));
-        })
-        .catch((error) => {
-          throw error;
-        });
+  Promise
+    .all([
+      new Promise(Promises.createKeypair),
+      new Promise(Promises.readOAuth2ClientSQLFile)
+    ])
+    .then((results) => {
+      let keypair = results[0];
+      let sql = results[1];
+      return Promise.all([
+        erp.execute(new Query(sql, [
+          script.iss,
+          script.iss,
+          forge.pki.publicKeyToPem(keypair.publicKey)
+        ])),
+        new Promise(Promises.saveP12File(keypair.privateKey))
+      ]);
     })
     .then(() => {
       return erp.execute(new Query([
--- a/scripts/keygen.js
+++ b/scripts/keygen.js
@@ -52,10 +52,7 @@
       for (let parameter in parameters) {
         if (parameters.hasOwnProperty(parameter)) {
           query = query.replace(
-            new RegExp(
-              '{{ parameter }}'.replace('parameter', parameter),
-              'g'
-            ),
+            new RegExp(`{{ ${parameter} }}`, 'g'),
             parameters[parameter]
           );
         }
@@ -65,7 +62,6 @@
   }
 
   let erp = new DataSource();
-
   Promise.resolve()
     .then(() => {
       return new Promise((resolve, reject) => {
--- a/scripts/keygen.js
+++ b/scripts/keygen.js
@@ -23,20 +23,26 @@
      */
     this.execute = function (query) {
       return new Promise((resolve, reject) => {
-        this.dataSource.query(query.sql(), this.credentials, function (error, result) {
-          if (error) {
-            reject(error);
+        this.dataSource.query(
+          query.sql(),
+          Object.assign(this.credentials, {
+            parameters: query.parameters()
+          }),
+          function (error, result) {
+            if (error) {
+              reject(error);
+            }
+            resolve(result);
           }
-          resolve(result);
-        });
+        );
       });
     };
   };
 
   /**
    * @constructor
-   * @param {(string|string[])}      query
-   * @param {Object.<string,string>} parameters
+   * @param {(string|string[])} query
+   * @param {Array.<string>}    parameters
    */
   let Query = function (query, parameters) {
     /**
@@ -46,15 +52,13 @@
       if (Array.isArray(query)) {
         query = query.join(' ');
       }
-      for (let parameter in parameters) {
-        if (parameters.hasOwnProperty(parameter)) {
-          query = query.replace(
-            new RegExp(`{{ ${parameter} }}`, 'g'),
-            parameters[parameter]
-          );
-        }
-      }
       return query;
+    };
+    /**
+     * @returns {Array.<string>}
+     */
+    this.parameters = function () {
+      return parameters;
     };
   };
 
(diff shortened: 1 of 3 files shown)


scripts/keygen.js
```diff
@@ -1,26 +1,98 @@
...
+                if (error) {
+                  throw error;
+                }
+              }
+            );
+          }
+        );
       }
-    });
+    );
```
bendiy: This would be much easier to follow if it was rewritten to use Promises. You're 4 callbacks deep here.

The `xdruple` schema isn't in core, so we should consider putting this script under the `xdruple-extension` repo. If it was moved to `xdruple-extension`, `promiseQuery()` would be available to use.
https://github.com/xtuple/xdruple-extension/blob/1_1_x/node-datasource/lib/promiseQuery.js

The `fs` functions can be promisified (`Promise.promisifyAll(fs);)` like this:
https://github.com/xtuple/xdruple-extension/blob/1_1_x/node-datasource/resources/FileResource.js#L32

Which gives you `fs.readFileAsync()` and `fs.writeFileAsync()`.

`forge.pki.rsa.generateKeyPair()` can probably be promisified the same way since its last parameter is a callback.

If we choose not to use Promises, our typical callback pattern is to not nest them. Instead, define each function at the top level scope and then pass the defined functions in as parameters.

scripts/sql/application.sql
```diff
@@ -0,0 +1,5 @@
+INSERT INTO xdruple.xd_site (
+  xd_site_name,
+  xd_site_url
+)
+VALUES ('{name}', '{url}');
```
bendiy: Unless these query files are going to be used elsewhere, I don't think there's much value in separating them into their own files vs. embedding the SQL directly in the `keygen.js` source.

amikheychik: The application.sql is ok to keep in the JS code, but oauth2 insert is too large to inlining.

scripts/keygen.js
amikheychik: If we move it to the xdruple-extension, it would add `commander` dependency there too. This script has a very narrow usage, so it should be fine to keep it in the xtuple while we're removing the mobile client code.

I'll look into promises.

gilmoskowitz: Please don't add another dependency to the xtuple repository. If the version of node we're using can handle Promises directly then that's fine. If we would have to add a package to get them, stick with callbacks (and break them out as Ben suggested).

amikheychik: @gilmoskowitz The Bluebird package that provides Promises is already a dependency for `xdruple-extension`, but moving script there would add `commander` as a dependency for `xdruple-extension`. The best way to resolve this, would be to switch from "parallel" packaging into "tree" packaging, when `xdruple-extension` would require `xtuple` package (and `xdruple-extension` itself would be required by the project package).

It looks like we can try to utilize npm itself for that (instead of Composer): private packages are available for $7/month (https://www.npmjs.com/pricing).

@bendiy extracting callbacks makes code less readable, as each callback function is used once, so in addition to the callback hell we add a code reading `goto` hell (instead of reading code one layer after another, it would require to jump around the code.

I'll keep experimenting with object-oriented JS to see if I can achieve some encapsulation here, to simplify the process.

amikheychik:
@gilmoskowitz @bendiy don't merge the current state yet. It's working, but with the last two commits I started on removing code duplication. I will either get to the state, when readability is good enough without promises, or just implementing simple promises in the script.

@bendiy side question, is it safe to use ECMAScript 6 `(params) => {}` short notation for anonymous functions? Looks like it's supported.
@gilmoskowitz would you have objections to switch to a shorter and cleaner syntax?

scripts/keygen.js
amikheychik: @gilmoskowitz  According to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) are supported in Node since v0.12. Is that correct, @bendiy ?

gilmoskowitz:
@amikheychik 

- The shorter syntax is ok for very short functions. The relative savings (a couple of words?) for longer functions is not worth the potential confusion. i haven't read enough examples yet to know where the threshold is between "very short" and "longer".
- Regarding promises and dependencies
  - xdruple-extension may depend on Bluebird now but we should remove that dependency, not push it further down the stack
  - We've had lots of trouble dealing with packages for a number of reasons (some technical, some not), so I'm very hesitant to increase our reliance on npm

amikheychik:
@gilmoskowitz I'm on board with zero (outside) dependencies concept. It looks like we can just use default ECMAScript 6 Promises here.

amikheychik:
@gilmoskowitz turns out, `() => {}` syntax is not mere sugar:

```
   this.execute = function (query, success) {
      this.dataSource.query(query.sql(), this.credentials, function (error) {
        console.log(arguments);
      });
    };
```

Arguments would be for `function (error) {}` callback, but

```
   this.execute = function (query, success) {
      this.dataSource.query(query.sql(), this.credentials, (error) => {
        console.log(arguments);
      });
    };
```

arguments are for `function (query, success) {}`

amikheychik:
@bendiy @gilmoskowitz code is good to be reviewed again

scripts/keygen.js
```diff
@@ -1,26 +1,140 @@
...
+    this.dataSource = require('../node-datasource/lib/ext/datasource').dataSource;
+    this.credentials = require(script.config).databaseServer;
+    this.credentials.database = script.database;
+    /**
+     * @param {Query} query
+     * @return {Promise}
+     */
+    this.execute = function (query) {
+      return new Promise((resolve, reject) => {
+        this.dataSource.query(query.sql(), this.credentials, function (error, result) {
```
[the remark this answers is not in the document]
amikheychik: Ah, I didn't know (or have already forgotten) that there is parameters handling in `DataSource.query`. Though, `Query()` constructor would still be useful to pair query and parameters. Same as in `php-util` [Query interface](https://github.com/xtuple/php-util/blob/master/src/Xtuple/Util/Postgres/Query/Query.php).

scripts/keygen.js
```diff
@@ -1,26 +1,140 @@
+#!/usr/bin/env node
+/* jshint node:true, esversion:6 */
```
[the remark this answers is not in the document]
amikheychik: I support switching to using this file, but this would require removing comments out of every file it was used and adding `.jshintrc` into every package we have. Semantically though, this is a not a part of this pull request, but I can easily do this transition as my next step.

scripts/keygen.js
```diff
@@ -1,26 +1,140 @@
...
-    fs.writeFile('private.p12', new Buffer(new Buffer(
-      forge.asn1.toDer(
-        forge.pkcs12.toPkcs12Asn1(keypair.privateKey, null, 'notasecret')
-      ).getBytes(),
-      'binary'
-    ), 'base64'), function (error) {
-      if (error) {
-        console.log(error);
+      for (let parameter in parameters) {
+        if (parameters.hasOwnProperty(parameter)) {
```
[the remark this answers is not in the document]
amikheychik: That's a good pattern. I'm still catching up on the current state of JS and all the improvements in the language since ECMAScript 6 and though `forEach` still exists only in jQuery :)

amikheychik:
@bendiy I address your comments, plus achieved better code logic separation and got rid of some duplication.

scripts/keygen.js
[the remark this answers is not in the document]
gilmoskowitz: In this case the in-file jshint is appropriate because it's different from other JavaScript files in the repository. We've already started using `.jshintrc` in the test subdirectory and should expand its use to the top level of the repository once we've removed the deprecated web client code. Otherwise it's just going to generate a lot of noise.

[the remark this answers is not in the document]
amikheychik: @gilmoskowitz not sure why would it be noise to remove all comments in one commit replaced with `.jshintrc.` file. The only question for me was to do it before or after removing mobile client code, and now I know the answer: deprecated code to be removed first.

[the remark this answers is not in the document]
gilmoskowitz: Definitely remove the deprecated code first. The noise I was referring to is from jshint failures, not commit diffs.

PR #3461, opened 2018-10-20
Recurring sales order

interhawk:
Database changes to enable creating recurring sales orders


foundation-database/public/functions/copysoheader.sql
```diff
@@ -0,0 +1,381 @@
...
+  END IF;
+
+  INSERT INTO charass
+        (charass_target_type, charass_target_id,
+         charass_char_id, charass_value)
+  SELECT charass_target_type, _soheadid,
+         charass_char_id, charass_value
+    FROM charass
+   WHERE ((charass_target_type='SO')
+     AND  (charass_target_id=pSoheadid));
```
andersdd: Don't know the answer to this (@gilmoskowitz) but should we also copy the document assignments in `docass`?

foundation-database/public/patches/populate_recur.sql
```diff
@@ -0,0 +1,6 @@
+INSERT INTO recurtype (recurtype_type, recurtype_table, recurtype_donecheck, 
+                       recurtype_schedcol, recurtype_limit, recurtype_copyfunc, 
+                       recurtype_copyargs, recurtype_delfunc) 
+               VALUES ('SALESORDER', 'cohead', 'isRecurSODone(cohead_id)', 'cohead_orderdate',
```
andersdd: Should `'SALESORDER'` actually be `'S'` ?

foundation-database/public/trigger_functions/cohead.sql
```diff
@@ -651,16 +651,41 @@ CREATE TRIGGER coheadBeforeDeleteTrigger
...
   DELETE FROM charass
   WHERE charass_target_type = 'SO'
     AND charass_target_id = OLD.cohead_id;
 
+  --recurrence cleanup
+  SELECT recur_id INTO _recurid
+    FROM recur
+   WHERE ((recur_parent_id = OLD.cohead_id)
+     AND  (recur_parent_type = "SALESORDER"));
+  IF (_recurid IS NOT NULL) THEN -- The deleted SO is the parent of a recurrence series
```
andersdd: `SALESORDER` should be switched to 'S'.  Also watch the xTuple coding standards when it comes to excessive brackets in WHERE clauses

foundation-database/public/functions/copysoheader.sql
interhawk: the new copysoheader procedure is simply a portion of the existing copyso procedure that was pulled out into a separate function to enable copying a sales order without copying line items. Copying document assignments may well be a good thing to consider, but  I think outside the scope of this pr maybe?

foundation-database/public/patches/populate_recur.sql
interhawk: Good catch, done.

foundation-database/public/trigger_functions/cohead.sql
interhawk: implemented

foundation-database/public/functions/copysoheader.sql
gilmoskowitz: I agree with @interhawk — out of scope. Copying `docass` is fraught with danger, as some documents might contain information specific to the original order (e.g. signed contracts) while others might be boilerplate.

PR #3465, opened 2018-10-23
[33969] Update Contact Merge for new contact fields

andersdd:
Could not reproduce exact error in issue but found newly added contact fields were causing problems.  Contact merge works without SQL error after this update


srigurudatta:
Performed UAT: Issue is fixed. User is able to merge contacts successfully. Please find the attached screenshot.
image

PR #3469, opened 2018-10-24
[34055] Resolve partial save trigger SQL errors on new Incident

andersdd:
Data integrity checks still exist in cpp code but had to be removed from the Trigger when partially saving an incident.


lalithapadala:
Performed UAT: User is fixed. 'Error Saving Incident' dialog is not displayed while creating or editing the incident category.

PR #3471, opened 2018-10-29
[34077] Resolve Sales Rep deletion

andersdd:
Trigger was returning an incorrect NEW (thought this would cause an error, but it was silent)


lalithapadala:
Performed UAT: Issue is fixed. Sales Representative is deleted successfully. Please find the attached screenshots

PR #3481, opened 2018-11-04
[34122] Fix saving of Employee when no image is selected

andersdd:
caused by 4_12 merge issue


lalithapadala:
Performed UAT: Issue is fixed. Employee is created successfully without any error when no image is selected.

PR #3485, opened 2018-11-07
[33376] added check to postVoucher

mmicr001:
Check if voucher has an item with no tagged receipts or has an item with no distributions.
If so raise exception.

Also added translations for all exceptions in storedProcErrorLookup.cpp

### DIFF
--- a/foundation-database/public/functions/postvoucher.sql
+++ b/foundation-database/public/functions/postvoucher.sql
@@ -66,6 +66,22 @@ BEGIN
 
   _glDate := COALESCE(_p.vohead_gldistdate, _p.vohead_distdate);
 
+--  Disallow posting voucher if it has any voucherItems without any tagged receipts or distributions
+  IF(EXISTS(SELECT 1 
+              FROM voitem 
+             WHERE voitem_vohead_id = pVoheadId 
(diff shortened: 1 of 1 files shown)


foundation-database/public/functions/postvoucher.sql
```diff
@@ -66,6 +66,22 @@ BEGIN
...
+             WHERE voitem_vohead_id = pVoheadId 
+               AND (NOT EXISTS(SELECT 1 
+                                FROM recv 
+                               WHERE recv_voitem_id = voitem_id 
+                                 AND recv_vohead_id = voitem_vohead_id) 
+                    OR NOT EXISTS(SELECT 1 
+                                   FROM vodist 
+                                  WHERE vodist_vohead_id = voitem_vohead_id 
+                                    AND vodist_poitem_id = voitem_poitem_id))))
+    THEN RAISE EXCEPTION 'Cannot post Voucher #% as it has items without any tagged receipts or without distributions [xtuple: postVoucher, -11, %]',
```
ChristopherCotnoir: Move the THEN to the previous line to be consistent with other code (see the above IF). Also put a space between the IF and ( while you're at it.

PR #3490, opened 2018-11-13
Nightly package build

gilmoskowitz:
- increase the number of extensions automatically loaded into the databases
- flaw: always builds the extension .gz files, which may not be desired for true release builds of reference databases


ChristopherCotnoir:
jenkins - retest this please

PR #3491, opened 2018-11-15
[34150] In ‘Copy Item’ screen, while copying the Item ‘Error Copying Item’ error dialog is displayed.

mmicr001:
The bug (as far as I can tell) is the result of the postbooks DB being initialized to a state that causes an error when rows are added to the `urlinfo` table.

This is because (after a fresh postbooks DB restore): 
- `urlinfo` starts off with 1 row, with `url_id = 4`
- the `url_id` is generated by the sequence `urlinfo_url_id_seq` which starts at 2 
Therefore, at some point as rows are added to `urlinfo` we will try to insert a row with `url_id=4` when there is already one in the table, causing the error message.
This is also the reason I was able to reproduce the error dialog, by attaching a WebSite to a incident's document.

I noticed during my testing that `url_id` and `file_id` were not following the same sequence, because the `createFile` and `createUrl` functions would call nextval() on their respective sequences. That is what this code addresses.

### DIFF
--- a/foundation-database/public/functions/createfile.sql
+++ b/foundation-database/public/functions/createfile.sql
@@ -3,8 +3,20 @@ drop function if exists createFile(TEXT, TEXT, BYTEA) cascade;
 create or replace function createFile(pTitle TEXT, pDescription TEXT, pStream BYTEA, pMimeType TEXT DEFAULT NULL) returns integer as $$
 declare
   _id integer;
+  pDefault text;
 begin
-  _id := nextval('file_file_id_seq');
+  select column_default 
+    into pDefault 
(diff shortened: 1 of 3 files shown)


mmicr001:
[This line](https://github.com/xtuple/xtuple/blob/ee060c440712bb30b0ab017c47eb3e95a4cf1ff9/foundation-database/public/tables/urlinfo.sql#L31) is always true, which I guess is ok since it changes the `urlinfo`'s sequence to `file`'s [sequence](https://github.com/xtuple/xtuple/blob/ee060c440712bb30b0ab017c47eb3e95a4cf1ff9/foundation-database/public/tables/urlinfo.sql#L37).

PR #3492, opened 2018-11-16
[33126] Currency edits not updating the currency ComboBox in Setup

andersdd:
Trigger notification was only refreshing the first XComboBox to use that notification

Added a secondary notification trigger that can be called by specific tables that need this (only applied to curr_symbol so far)


andersdd:
Closing as a simpler solution has been found

PR #3505, opened 2018-11-26
Normalize country names in demo databases

ChristopherCotnoir:


ChristopherCotnoir:
The demo data has some issues. All the 'USA' are actually in the state columns instead of country because one of the addresses somehow had most of its columns shifted by one and then it was propogated throughout. I can confirm that there are know country names in any country fields after a build.

PR #3528, opened 2018-12-10
[34280] Resolve PO Return Credit Memo error

andersdd:


foundation-database/public/functions/postporeturncreditmemo.sql
```diff
@@ -137,7 +137,7 @@ BEGIN
                                 getMiscTaxtypeId(), FALSE, ARRAY[''],
                                 ARRAY[_p.poitem_taxtype_id],
                                 ARRAY[_itemAmount])->'lines'->0->'tax')
-      GROUP BY (value->>'taxid')::INTEGER
+      GROUP BY jsonb_array_elements.value, (value->>'taxid')::INTEGER
```
ChristopherCotnoir: Adding this will cause it to not group different value with the same value->>'taxid' together anymore. Unfortunately due to postgres's apparent poor handling of json the json elements will have to be selected in a subquery and then grouped a level up in order to get the intended behavior.

andersdd: Are you sure? From what I could see the calculatetax() function returns a single JSONB row in which case the GROUP BY effectively does nothing.

srigurudatta:
Performed UAT: Issue is fixed. User is able to create Credit Memo for uninvoiced PO Return. Please find the attached screenshot.
poreturn_cm

foundation-database/public/functions/postporeturncreditmemo.sql
andersdd: I still think it works.  Here is an example of a Tax Zone with multiple tax codes assigned.  This is the result which feeds into the loop.
image

gilmoskowitz: Chris is right. The only member of `value` that's being aggregated is `value->>'tax'` so the other members need to be aggregated or added to the `group by`:

```SQL
select count(*), value->>'b', sum((value->>'v')::integer)
  from jsonb_array_elements('[ { "a": 1, "b": true, "v": 1 }, { "a": 2, "b": true, "v": 2 }, { "a": 1, "b": true, "v": 3 } ]'::jsonb)
 group by value->>'a';
ERROR:  column "jsonb_array_elements.value" must appear in the GROUP BY clause or be used in an aggregate function
LINE 1: select count(*), value->>'b', sum((value->>'v')::integer)
                         ^
select count(*), max(value->>'b'), sum((value->>'v')::integer)
  from jsonb_array_elements('[ { "a": 1, "b": true, "v": 1 }, { "a": 2, "b": true, "v": 2 }, { "a": 1, "b": true, "v": 3 } ]'::jsonb)
 group by value->>'a';
 count | max  | sum
-------+------+-----
     1 | true |   2
     2 | true |   4
(2 rows)

select count(*), value->>'b', sum((value->>'v')::integer)
  from jsonb_array_elements('[ { "a": 1, "b": true, "v": 1 }, { "a": 2, "b": true, "v": 2 }, { "a": 1, "b": true, "v": 3 } ]'::jsonb)
 group by value->>'a', value->>'b';
 count | ?column? | sum
-------+----------+-----
     1 | true     |   2
     2 | true     |   4
(2 rows)
```

ChristopherCotnoir: Dave is right, actually, the GROUP BY can be removed. It's there because there was one there before I modified this query, but it seems that was a mistake. The idea is to group together multiple instances of the same tax on the line. It is possible to have that by having a sub tax assigned to the zone/type pair and have its basis tax assigned to the zone/type pair, but I'm pretty sure that's not supposed to be possible as you have to do things in a specific order to sneak it past the application and it doesn't really make any sense. Other than that, the same tax code will never appear twice on the same line.

PR #3531, opened 2018-12-11
[33322] Convert some CRM metaSQL over to generic search method

andersdd:
Issue requirements related to CRM queries.  Plus this PR provides some examples of its usage.

requires xtuple/qt-client#2195 before this change will work.

### DIFF
--- a/foundation-database/public/tables/metasql/prospects-detail.mql
+++ b/foundation-database/public/tables/metasql/prospects-detail.mql
@@ -104,12 +104,10 @@ WHERE true
 <? literal("charClause") ?>
 )
 SELECT * FROM _prospects
-WHERE true
 <? if exists("search_pattern") ?>
-  AND ( FALSE
+  WHERE FALSE
   <? foreach("search_fields") ?>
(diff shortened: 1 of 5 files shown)


foundation-database/public/tables/metasql/prospects-detail.mql
```diff
@@ -113,4 +102,14 @@ WHERE true
...
+)
+SELECT * FROM _prospects
+WHERE true
+<? if exists("search_pattern") ?>
+  AND ( FALSE
+  <? foreach("search_fields") ?>
+    OR <? literal("search_fields") ?>::TEXT ~* <? value("search_pattern") ?>
+  <? endforeach ?>
+)
+<? endif ?>
```
gilmoskowitz: These can all be simplified a bit more:

```
SELECT * FROM _cte
<? if exists("search_pattern") ?>
 WHERE FALSE
   <? foreach("search_fields") ?>
    OR <? literal("search_fields") ?>::TEXT ~* <? value("search_pattern") ?>
   <? endforeach ?>
<? endif ?>
```

We've seen `WHERE true` slow down some queries but we don't know if the marginal improvement from removing it here is worth the effort.

PR #3536, opened 2018-12-18
[34312] update tables with fkeys to prjtask and todoitem allows updating xtte

gilmoskowitz:
the time and expense extension won't upgrade properly without this
because the teitem records would point to non-existent or incorrect
task records

### DIFF
--- a/foundation-database/public/patches/task_migration.sql
+++ b/foundation-database/public/patches/task_migration.sql
@@ -114,6 +114,29 @@ IF EXISTS (SELECT 1
     FROM prjtaskmap
    WHERE charass_target_type = 'TASK' AND charass_target_id = prjtask_id;
 
+  FOR _rec IN SELECT fn.nspname, ft.relname, attname, conname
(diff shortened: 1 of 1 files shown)


gilmoskowitz:
@andersdd I think `teitem` is not being updated correctly. It looks like the f-key is [dropped](https://github.com/xtuple/xtte/blob/2_5_x/extensions/time_expense/foundation-database/te/tables/teitem.sql#L78) shortly after being [created](https://github.com/xtuple/xtte/blob/2_5_x/extensions/time_expense/foundation-database/te/tables/teitem.sql#L73). Did I miss something important?

andersdd:
@gilmoskowitz  I agree `teitem` does not look correct.  TBH it looks like a git merge problem.  The Drop Constraint should be above the recreation of the constraint.    The problem will lie in the timing of upgrade steps.  Does the `migrate_tasks` run before the `xtte` package is installed and will that prevent the migrate from running?

PR #3537, opened 2018-12-20
[32398] Freight doubling under specific circumstances

mmicr001:
This change only affects the freight calculations when the setup has been configured to not automatically calculate freight. 
The bug is caused by the `shiphead_freight` for all shipments being equal to the total SO freight amount (`cohead_freight`).
Its not ideal, but it adds all the freight from the SO to the first shipment (`shiphead_freight`) and 0 to the rest of the shipments.


srigurudatta:
Performed UAT: Tested with the given scenario for the issue #32398 and observed that Issue is fixed. Please refer to the attached screenshot.
so_invoice_freight

ChristopherCotnoir:
This fix should be fine, but may not be complete. Looking at shipOrder.cpp, the freight can get overridden several ways, which might account for the scenarios with a single site that the customer mentioned.

PR #3559, opened 2019-01-25
[34337] fix access errors caused by missing role grants

gilmoskowitz:


gilmoskowitz:
jenkins - retest this please
you might not be merging pull requests before starting the tests

ChristopherCotnoir:
Have you confirmed this actually works? These grants should be being run by fixacl(), so if that isn't working I don't see why this would. My suspicion is the creation of the temp tables in migrate_tax is the culprit, in which case these statements (or fixacl) have to be run after migrate_tax.

gilmoskowitz:
Confirmed this works by upgrading a 4.11.3 database to the latest on this branch.

PR #3560, opened 2019-02-01
Make calculatetax function more robust for when no freight is required

andersdd:
Function returned NULL if no freight taxtype id was supplied in params.  Found this when attempting to utilise via another function


ChristopherCotnoir:
Not sure if you still need this, but this is probably not the best fix as it will still return the freight in the resulting json which means it will be saved with a value of 0 when what you are really trying to say is that there is no freight. Would be more appropriate to handle it like when misc is null, see _includemisc.

andersdd:
Updated as suggested

PR #3566, opened 2019-02-12
[34341] fix idempotence issue unlocking in web-enabled dbs

gilmoskowitz:


database/source/xt/functions/pg_advisory_unlock.sql
```diff
@@ -1,14 +1,15 @@
...
+  RETURNS BOOLEAN AS $$
+-- Copyright (c) 1999-2019 by OpenMFG LLC, d/b/a xTuple. 
+-- See www.xtuple.com/CPAL for the full text of the software license.
+-- simplified version of lib/orm/source/xt/javascript/data.sql::releaseLock()
+BEGIN
+  PERFORM pg_catalog.pg_advisory_unlock(pOid, pId);
+  DELETE FROM xt.lock
+   WHERE lock_table_oid = pOid
+     AND lock_record_id = pId
+     AND lock_username  = geteffectivextuser();
```
ChristopherCotnoir: Should also filter on lock_pid as above.

database/source/xt/functions/pg_advisory_unlock.sql
```diff
@@ -1,14 +1,16 @@
...
+-- See www.xtuple.com/CPAL for the full text of the software license.
+-- simplified version of lib/orm/source/xt/javascript/data.sql::releaseLock()
+BEGIN
+  PERFORM pg_catalog.pg_advisory_unlock(pOid, pId);
+  DELETE FROM xt.lock
+   WHERE lock_table_oid = pOid
+     AND lock_record_id = pId
+     AND lock_username  = geteffectivextuser()
+     AND lock_pid       = pg_backend_pid();
+  RETURN TRUE;
```
ChristopherCotnoir: I understand we're making this change to avoid bugging the user when something goes wrong, but it's worth noting that the core postgres function we are overriding also returns false when trying to release a lock not held. It might be worth preserving that behavior here since we are using the same function name, and maybe change the code in qt-client if we really don't want to do anything if it tries to release a non-existent lock. Just something for consideration.

PR #3569, opened 2019-02-20
[34566] Remove references to todos

ChristopherCotnoir:


ChristopherCotnoir:
Requires xtuple/qt-client#2229 and xtuple/xtdesktop#112, associated with xtuple/connect#136.

gilmoskowitz:
This group of pull requests goes much further than the original bug report, which was purely textual in nature -- fix labels on windows. That's not necessarily a problem; it just means we need to review this much more carefully than anticipated.

foundation-database/public/patches/task_migration.sql
```diff
@@ -235,15 +236,18 @@ IF EXISTS (SELECT 1
                   NATURAL JOIN todotaskmap
                  WHERE todoitem_username IS NOT NULL;
 
-  UPDATE alarm SET alarm_source_id = task_id
+  UPDATE alarm SET alarm_source = 'TASK',
+                   alarm_source_id = task_id
     FROM todotaskmap
    WHERE alarm_source = 'TODO' AND alarm_source_id = todoitem_id;
```
andersdd: There is the potential for duplicated keys with this migration.  Between the legacy 'J' and 'TODO' sources with the same task_id.

ChristopherCotnoir:
The reason I merged the sources is because some of the fields in the source table are eventually shown to the user. That's what it's for. I could have maybe had two identical source entries instead, but that seemed like it would have its own problems.

foundation-database/public/patches/task_migration.sql
ChristopherCotnoir: Hmm, seems like this migration script already had a problem then. Because if that's the case, as it stands we already have a duplication, just to two different sources that actually represent the same thing, but the user would still see the task twice.

ChristopherCotnoir:
Accidentally pushed branch to xtuple instead of my fork. Replaced by xtuple/xtuple#3573.

PR #3571, opened 2019-02-21
[34298] Fix references to non-existent column

ChristopherCotnoir:


harithamekala:
Hi Chris,
We've observed that Account Merge issue is doesn't exists for 4.12.0 builds, but which exists in 5.0 Beta build. Please give the fix to 5.0

gilmoskowitz:
@harithamekala This pull request is explicitly for the 4_12_x branch. The fix will be available in 5_0_x next time we do a code merge, which will happen as part of the 5.0.0-beta release preparation.

PR #3570, opened 2019-02-21
[34597] Finish country migration automatically when possible

ChristopherCotnoir:

PR #3572, opened 2019-02-25
[34597] fix test failures

gilmoskowitz:
merge after or possibly instead of https://github.com/xtuple/xtuple/pull/3570
not sure how CI could have passed before now as lots of tests are failing, sometimes with product bugs and sometimes with test bugs.


gilmoskowitz:
Unnecessary after #3570 was fixed to enable triggers again. The imageass trigger re-enabling will be handled in a separate pull request.

PR #3574, opened 2019-02-26
fix imageass trigger enabling bug found while testing other things

gilmoskowitz:


ChristopherCotnoir:
Good to merge once CI passes.

PR #3577, opened 2019-02-28
[18464] Fix issues with API armemo including sql error and paid amount

andersdd:
Amended API view to match the underlying insert function parameters.  Tested and confirmed api.armemo is saving the INSERTed memos correctly.

Removed `paid` column from the api view as this was misleading and incorrect.  See associated issue for a more detailed explanation.

As this change involved an amendment to the API structure/view have posted to version 5.0.


gilmoskowitz:
replaced with #3580

PR #3592, opened 2019-03-12
[34164] Debit the correct G/L account when crediting a cash receipt

ChristopherCotnoir:


ChristopherCotnoir:
Associated with xtuple/qt-client#2263, requires xtuple/payment-gateways#63.

PR #3593, opened 2019-03-13
[33844] Add missing Documents report and tidy up multi-tier layout

andersdd:
Can be upmerged to 5.0


srigurudatta:
Performed UAT: Issue is fixed. In ‘Documents’ screen,'Preview','Print' and 'Schedule' buttons functionality is working as expected.
Preview_Button
Schedule_screen
Print_ButtonScreen

PR #3600, opened 2019-03-18
Fix bugs found during demo

ChristopherCotnoir:


ChristopherCotnoir:
Associated with xtuple/qt-client#2273.

PR #3604, opened 2019-03-24
[23942] Add basic email validation and prep for external service

andersdd:


foundation-database/public/functions/validatemailaddress.sql
```diff
@@ -0,0 +1,14 @@
+CREATE OR REPLACE FUNCTION validateEmailAddress(pEmail TEXT)
```
gilmoskowitz: CI is failing because the filename doesn't match the function name (missing the `e` in `email`) but the manifest file does match (has the `ee`).

ChristopherCotnoir:
It's not the capitalization, as Gil said above it's the filename itself (1 e vs 2 e).

andersdd:
File has been renamed.

PR #3606, opened 2019-03-26
[34015] Revise ledger control metasql

ChristopherCotnoir:


ChristopherCotnoir:
Associated with xtuple/xtprjaccnt#49.

PR #3601, opened 2019-03-21
[28912] Fix characteristic query performance issue

ChristopherCotnoir:

PR #3607, opened 2019-03-27
[34786] Resolve characteristic issues with new Prospects

andersdd:
Resolves issue where multiple assignment of the same characteristic to a prospect resulted in duplicated rows in the prospects list

### DIFF
--- a/foundation-database/public/tables/metasql/prospects-detail.mql
+++ b/foundation-database/public/tables/metasql/prospects-detail.mql
@@ -5,8 +5,8 @@
 -- See www.xtuple.com/CPAL for the full text of the software license.
 
 WITH _prospects AS (
-SELECT prospect_id, prospect_crmacct_id,
-  prospect.*, prospect_created::DATE AS created,
+SELECT prospect_id AS id, prospect_crmacct_id AS altid,
+  prospect.*, prospect_created::DATE AS create,
   opsource_name AS prospect_source,
   cntct.*, getcontactphone(cntct_id, 'Office') AS contact_phone,
   addr.*
@@ -103,11 +103,12 @@ WHERE true
 <? endif ?>
 <? literal("charClause") ?>
 )
-SELECT * FROM _prospects
+SELECT DISTINCT ON (id) * FROM _prospects
 <? if exists("search_pattern") ?>
   WHERE FALSE
   <? foreach("search_fields") ?>
     OR <? literal("search_fields") ?>::TEXT ~* <? value("search_pattern") ?>
   <? endforeach ?>
 <? endif ?>
-ORDER BY prospect_number; 
+ORDER BY id;
+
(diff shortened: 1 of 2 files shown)


foundation-database/public/tables/metasql/prospects-detail.mql
```diff
@@ -5,8 +5,8 @@
 -- See www.xtuple.com/CPAL for the full text of the software license.
 
 WITH _prospects AS (
-SELECT prospect_id, prospect_crmacct_id,
-  prospect.*, prospect_created::DATE AS created,
+SELECT prospect_id AS id, prospect_crmacct_id AS altid,
+  prospect.*, prospect_created::DATE AS create,
```
gilmoskowitz: Dropping the `d` from `created` looks incorrect.

foundation-database/public/tables/metasql/prospects-detail.mql
```diff
@@ -103,11 +103,12 @@ WHERE true
 <? endif ?>
 <? literal("charClause") ?>
 )
-SELECT * FROM _prospects
+SELECT DISTINCT ON (id) * FROM _prospects
```
gilmoskowitz: Please don't add the `DISTINCT ON (id)` here. It masks a long-standing bug in the handling of characteristics. See https://github.com/xtuple/xtuple/pull/3601 for examples on the alternative, which both handles the duplicate rows problem and improves query performance when there are multiple characteristic assignments.

andersdd:
Revised query to better handle characteristics in the list - following current pattern.
Also slipped in cleanup of characteristic and document assignments when deleting prospects (issue found when testing list)

srigurudatta:
Performed UAT: Issue still exists. User is able to add duplicate characteristics with same values to the prospect.
Prospect_DuplicateChars

gilmoskowitz:
@srigurudatta The behavior you observed is correct behavior. The proposed fix addresses a problem in the list of _prospects_, not the list of _characteristics_. Given the data in the screenshot above, the old version of the query would show the prospect PRO2 at least twice in the Prospects window — **Sales > Prospect > List** — while the new query should show PRO2 only once.

foundation-database/public/tables/metasql/prospects-detail.mql
```diff
@@ -103,11 +126,14 @@ WHERE true
...
+SELECT *
+FROM _prospects
 <? if exists("search_pattern") ?>
   WHERE FALSE
   <? foreach("search_fields") ?>
     OR <? literal("search_fields") ?>::TEXT ~* <? value("search_pattern") ?>
   <? endforeach ?>
 <? endif ?>
-ORDER BY prospect_number; 
+ORDER BY id;
```
gilmoskowitz:
```suggestion
ORDER BY prospect_number;
```
Otherwise the order of output on a report that might use this won't make sense; an unsorted XTreeWidget won't either.

srigurudatta:
Performed UAT: Issue is fixed. Duplicate prospects are not created while adding the same characteristics to the prospect.

PR #3611, opened 2019-03-31
[34761] Revert closing of job costed W/O to shipping transaction.  

andersdd:
Resolve COGS bug where a job costed item was not generating COGS entries due to not finding an appropriate item cost from the W/O.

This fix does _not_ address historic data which will be corrupted.

requires associated cpp client change

Also fixes issue #34800


gilmoskowitz:
jenkins - retest this please
you reported a full disk last time

PR #3612, opened 2019-04-01
[34801] Copy S/O was not copying the characteristic pricing over.

andersdd:
Also resolves the costing of copied sales orders (initially set to zero) as costing comes from the work order which has yet to process any items and therefore determine costs.

### DIFF
--- a/foundation-database/public/functions/copyso.sql
+++ b/foundation-database/public/functions/copyso.sql
@@ -98,9 +98,9 @@ BEGIN
     -- insert characteristics first so they can be copied to associated supply order
     INSERT INTO charass
           (charass_target_type, charass_target_id,
-           charass_char_id, charass_value)
+           charass_char_id, charass_value, charass_price)
     SELECT charass_target_type, _soitemid,
-           charass_char_id, charass_value
+           charass_char_id, charass_value, charass_price
       FROM charass
      WHERE ((charass_target_type='SI')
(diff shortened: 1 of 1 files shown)


foundation-database/public/functions/copyso.sql
```diff
@@ -63,8 +63,9 @@ BEGIN
       COALESCE(pSchedDate, _soitem.coitem_scheddate),
       _soitem.coitem_promdate,
       _soitem.coitem_qtyord,
-      CASE WHEN fetchMetricBool('WholesalePriceCosting') THEN (SELECT item_listcost FROM item
-                                                               WHERE item_id=_soitem.itemsite_item_id)
+      CASE WHEN _soitem.jobcosted THEN 0.00
```
gilmoskowitz: `0.00` doesn't look right. The desktop client uses `SELECT unitcost FROM itemCost(...);` if no supply work order exists yet.

andersdd: I discussed this with Professional Services and the comment was job costed items get their item costing only from the generated W/O.  When copying a S/O the new W/O has yet to generate any costs so using `itemcost()` is incorrect.  Furthermore if you manually create a S/O for a job costed item the costing is zero, so if you copy this manual S/O and apply std. costing to the copy you end up with a discrepancy between the two sales orders.  This was how I initially identified the issue.

PR #3617, opened 2019-04-05
[34846] Add use tax for expense transactions

ChristopherCotnoir:


ChristopherCotnoir:
Requires xtuple/qt-client#2298 and xtuple/private-extensions#1091.

PR #3618, opened 2019-04-08
Fix metasql files to only have one query

ChristopherCotnoir:


ChristopherCotnoir:
Requires xtuple/qt-client#2305.

PR #3621, opened 2019-04-11
[34877] Fix invalid invoice/sales credit data

ChristopherCotnoir:


ChristopherCotnoir:
Associated with xtuple/qt-client#2307.

PR #3638, opened 2019-05-06
[34989] Update pseudofkeys on crmaccount child objects

ChristopherCotnoir:


foundation-database/public/functions/merge2crmaccts.sql
```diff
@@ -122,8 +123,53 @@ BEGIN
       IF (_debug) THEN
         RAISE NOTICE 'Updating CRM relation %,  src: %,  tgt: %,  keepSrc: %', _crmtbl, _srcid, _tgtid, _keepSrc;
       END IF;
+      _crmtype := CASE _crmtbl WHEN 'prospect' THEN 'PSPCT'
```
gilmoskowitz: `_crmtype` should probably be a `source%ROWTYPE`, perhaps set by `SELECT * FROM source WHERE source_table = _crmtbl;` The problem is that the string describing the pseudo-fkey is different for different tables (dumb but true):
- `crmacct` is indicated by `CRMA`, `CRMA`, and `CRMACCT` (`source_name`, `source_docass`, `source_charass`)
- `cntct` is indicated by `T`, `T`, and `CNTCT` (note this conflicts with `taxauth` handling below, which should be using `TAXAUTH`)

Otherwise the `changePseudoFKeyPointers` below will hide data by setting the type columns to unexpected values.

foundation-database/public/functions/merge2crmaccts.sql
```diff
@@ -122,8 +123,55 @@ BEGIN
...
+              + changePseudoFKeyPointers('public', 'docass', 'docass_source_id',
+                                         _srcid, 'public', _crmtbl, _tgtid,
+                                         'docass_source_type', _source.source_docass, true)
+              + changePseudoFKeyPointers('public', 'docass', 'docass_target_id',
+                                         _srcid, 'public', _crmtbl, _tgtid,
+                                         'docass_target_type', _source.source_docass, true)
+              + changePseudoFKeyPointers('public', 'imageass', 'imageass_source_id',
+                                         _srcid, 'public', _crmtbl, _tgtid,
+                                         'imageass_source', _source.source_docass, true)
+              ;
```
gilmoskowitz: As @ChristopherCotnoir pointed out in an off-line conversation, there may be cases where `source_docass = ''`. Building `_result` as a sum made sense when this function only dealt with CRM accounts. It might not anymore.

Suggestion for here:
```
_tmp := changePseudoFKeyPointers(...);
IF _tmp > 0 AND _source.source_docass = '' THEN
  RAISE EXCEPTION 'Cannot merge % since % records will be abandoned [...]', ...;
END IF;
_result := _result + _tmp;
```
and update `source_docass` in some `source` records where it's currently `''`.

PR #3640, opened 2019-05-07
[35017] Remove inappropriate column

ChristopherCotnoir:


AnjaniYeleti:
Performed UAT: Issue is fixed. In ‘Purchase Order’ screen, user is able to edit the Purchase Order line item successfully without any error.

PR #3641, opened 2019-05-08
[35029] Use id instead of name, add user to json

ChristopherCotnoir:
Code from @gilmoskowitz.


AnjaniYeleti:
Tested in xTuple 5.0.0 RC latest nightly build(2019-05-08): Issue is fixed. While opening an existing CRM Account via new CRM Account, error dialog is not displayed.

PR #3644, opened 2019-05-16
[34848] Cleanup createpr(). Check for existing PR before creating it.

bendiy:
When a user does not have `OverrideTax` priv, the `coitem_taxtype_change` trigger causes a cascade of updates to coitem which ends up creating two PRs for one line on a order.

Similar to how `createpurchasetosale()` checks for an existing PO before creating one, make `createpr()` check for an existing PR before creating one. This prevents the duplication of PRs.

Also cleanup the `createpr()` functions to reduce the duplication and follow latest coding standards.

### DIFF
--- a/foundation-database/public/functions/createpr.sql
+++ b/foundation-database/public/functions/createpr.sql
@@ -6,7 +6,8 @@ CREATE OR REPLACE FUNCTION createPr(
   pDueDate DATE,
   pNotes TEXT,
   pOrderType CHARACTER(1),
-  pOrderId INTEGER
+  pOrderId INTEGER,
+  pProjId INTEGER DEFAULT NULL::INTEGER
 ) RETURNS INTEGER AS $$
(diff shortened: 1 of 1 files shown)


foundation-database/public/functions/createpr.sql
```diff
@@ -194,46 +140,47 @@ BEGIN
...
-    _parent.duedate, pParentNotes );
+  _prid := createPr(
+    _orderNumber,
+    _parent.itemsiteid,
+    validateOrderQty(_parent.itemsiteid, _parent.qty, TRUE),
+    _parent.duedate,
+    COALESCE(pParentNotes, _parent.notes),
+    pParentType,
+    pParentId
+  );
```
ChristopherCotnoir: The old query inserts _parent.prjid into pr_prj_id but the query called by this overload does not.

bendiy: Good catch. Fixed now.

PR #3650, opened 2019-05-21
[35118] Correct default priority setting when creating Projects 

andersdd:
from CRM entity.  Fixed initial "from Opportunity" but issue affected "from Incident" as well.


srigurudatta:
Performed UAT: Issue is fixed.User is able to create a project from an opportunity without any errors.

PR #3657, opened 2019-05-29
[35144] Delete Contacts attached to CRM entities -  fix sql error

andersdd:


srigurudatta:
Performed UAT: Issue is fixed. User is able to delete the detached contact from the Contact list and from the Prospect-contact list without any error.

PR #3659, opened 2019-05-30
Allow overriding address for expense transaction use tax

ChristopherCotnoir:


ChristopherCotnoir:
Requires xtuple/qt-client#2357 and xtuple/private-extensions#1099.

PR #3673, opened 2019-06-11
[35218] Don't JOIN on poitem

ChristopherCotnoir:


ChristopherCotnoir:
Requires xtuple/private-extensions#1105.

PR #3674, opened 2019-06-11
[35175] Let reversecashreceipt handle debit memo creation

ChristopherCotnoir:


ChristopherCotnoir:
Associated with xtuple/payment-gateways#76.

PR #3677, opened 2019-06-17
[35236] Allow deleting employee group with employees

ChristopherCotnoir:


ChristopherCotnoir:
Requires xtuple/qt-client#2393.

PR #3682, opened 2019-06-18
[35182] Use NULL when -1

ChristopherCotnoir:


srigurudatta:
Performed UAT: Issue is fixed. User is able to ship the Sales Order for Reference item type. Please find the attached screenshot.
SOReferenceItem_Shipped

PR #3683, opened 2019-06-18
[35185] Add back missing drop

ChristopherCotnoir:


AnjaniYeleti:
Tested in xTuple 4.12.0 release build(2019-05-02) & 5.0.0 latest nightly build(2019-06-20) with Manufacturing edition and observed that user is able to void the posted voucher in 'Payment Register' screen. Please find the attached screenshot
paymentRegister

PR #3685, opened 2019-06-20
[35263] Don't update shipment freight when sales order freight changes

ChristopherCotnoir:


srigurudatta:
Performed UAT: Issue is fixed. If the user edited the SO freight amount after partial shipment, the freight amount for second invoice created after partial shipment is same as that of first invoice.

PR #3689, opened 2019-06-26
[35226] Don't copy custpn for new customer

ChristopherCotnoir:


foundation-database/public/functions/copyso.sql
```diff
@@ -77,7 +77,12 @@ BEGIN
       FALSE,
       0.0,
       NULL,
-      _soitem.coitem_custpn,
+      CASE WHEN (SELECT cohead_cust_id = COALESCE(pCustomer, cohead_cust_id)
+                   FROM cohead
+                  WHERE cohead_id = pSoheadid)
+           THEN _soitem.coitem_custpn
+           ELSE ''
+       END,
```
gilmoskowitz: The copy should take into account [the target customer's item aliases](https://github.com/xtuple/qt-client/blob/5_0_x/guiclient/salesOrderItem.cpp#L2216-L2234).

PR #3691, opened 2019-07-01
[35182] [35245] Resolve issue to ship error for non-controlled items

andersdd:
This is the same fix applied to version 5.0 - just backporting to resolve a separately raised issue


lalithapadala:
Performed UAT:  Issue is fixed. Non-Controlled items are issued to shipping successfully without any error. Please find the attached screenshot

PR #3692, opened 2019-07-01
[35307] Copy Invoice function was missing Alternate revenue account 

andersdd:
in the invoice item data copy section


lalithapadala:
Performed UAT:  Issue is fixed. ‘Alternate Revenue Account’ is copied successfully to the recurring invoices. Please find the attached screenshot

